PR #7, opened 2021-10-08
Migrate TS and create sign up API

ketphan02:


package.json
```diff
@@ -10,17 +10,21 @@
...
   "dependencies": {
     "@next-auth/prisma-adapter": "^0.5.2-next.15",
-    "@prisma/client": "^3.1.1",
+    "@prisma/client": "^3.2.1",
+    "@types/react": "^17.0.27",
     "fs": "^0.0.1-security",
     "next": "11.1.2",
-    "next-auth": "^4.0.0",
-    "prisma": "^3.1.1",
+    "next-auth": "^4.0.0-beta.4",
```
Opeyem1a, line 17: We're using the beta version?

ketphan02: Yeah, the documentation guided this way

PR #9, opened 2021-10-10
List of Map Cards

ketphan02:


src/components/MapCard/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+const Arrow: React.FC = () => {
+	return (
+		<svg width="15" height="12" viewBox="0 0 15 12" fill="none" xmlns="http://www.w3.org/2000/svg">
+			<path d="M0.899684 5.22821C0.507143 5.22617 0.187273 5.54273 0.185237 5.93528C0.1832 6.32782 0.499766 6.64769 0.892308 6.64972L0.899684 5.22821ZM13.9117 6.5091C14.1907 6.23297 14.193 5.78294 13.9169 5.50394L9.4171 0.957218C9.14097 0.678208 8.69094 0.675873 8.41193 0.952002C8.13292 1.22813 8.13059 1.67816 8.40672 1.95717L12.4065 5.99869L8.36499 9.99849C8.08598 10.2746 8.08365 10.7247 8.35977 11.0037C8.6359 11.2827 9.08593 11.285 9.36494 11.0089L13.9117 6.5091ZM0.892308 6.64972L13.408 6.71467L13.4154 5.29315L0.899684 5.22821L0.892308 6.64972Z" fill="#F16B67"/>
+		</svg>
+	);
+}
+
+const ConditionalCard: React.FC<{ type: 'list' | 'grid' }> = ({ type }) => {
```
Opeyem1a, line 20: I'm kinda confused what conditional card does? So like this is the component that conditionally determines which buttons show up in a card? If so, it should be named more accordingly

src/components/MapCard/mapCard.module.scss
```diff
@@ -0,0 +1,77 @@
+$primary-color: #f16b67;
+ 
+.cardHeightNone {
```
Opeyem1a, line 3: 'None" implies 0 height, I think we should use the word 'default' in these cases

src/components/MapCard/mapCard.module.scss
```diff
@@ -0,0 +1,77 @@
+$primary-color: #f16b67;
```
Opeyem1a, line 1: Colour choices like this should be defined in the global scss files (which is my task, but for now just put it in the global file)

src/components/MapCard/mapCard.module.scss
```diff
@@ -0,0 +1,77 @@
...
+ 
+.cardHeightNone {
+    height: 168px;
+}
+
+.cardHeightButton {
+    height: 240px;
```
Opeyem1a, line 8: I'm generally a fan of using `rem` instead of `px` unless necessary

src/components/MapCard/mapCard.module.scss
```diff
@@ -0,0 +1,77 @@
...
+            font-size: 24px;
+            font-weight: 900;
+        }
+
+        .difficulty {
+            font-size: 14px;
+            font-weight: 800;
+            
+            .easy {
+                color: #05C05B;
```
Opeyem1a, line 38: These colours should be constants defined at the top of this file, though

src/components/MapCard/mapCard.module.scss
```diff
@@ -0,0 +1,77 @@
...
+        justify-content: space-evenly;
+        align-items: center;
+
+        // draw a square
+        .arrowBox {
+            width: 26.81px;
+            height: 26.81px;
+            left: 110px;
+            top: 143.7px;
+            background-color: #FFFFFF;
```
Opeyem1a, line 68: Make a constant in the global file named `$white` (since it might not literally be white later on for aesthetic/contrast reasons

src/components/MapCardList/index.tsx
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react';
```
Opeyem1a, line 1: I think we should still name the files the same name as the folder just so it's obvious when looking at the filename, unless that's bad practice?

src/components/MapCard/mapCard.module.scss
```diff
@@ -0,0 +1,77 @@
...
+            border-radius: 4.26456px;
```
Opeyem1a, line 69: Oh is this from the Figma file? I think we can round numbers to 2 decimal places going forward

src/components/MapCard/mapCard.module.scss
```diff
@@ -0,0 +1,77 @@
...
+        }
+    }
+
+    button {
+        width: 142px;
+        height: 40px;
+        border-radius: 6px;
+        border: none;
+        background-color: $primary-color;
+        color: white;
```
Opeyem1a, line 57: This should also be `$white`

src/components/MapCard/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+	if (type === 'list') {
+		return (
+			<button>
+				<p>Solve Online</p>
```
Opeyem1a, line 24: Does text inside of a button need a paragraph tag?

src/components/MapCard/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+					<Arrow />
+				</span>
+			</button>
+		)
+	}
+}
+
+const Card: React.FC<MapCardProps> = ({ title, desc, diff, type }) => {
+
+	const Diff = () => {
```
Opeyem1a, line 35: Do we need this? The only difference is the class name, couldn't we just have a function that returns the class name based on the difficulty and put the jsx into line 49 directly, calling that function for the classname?

Opeyem1a (changes requested):

next-env.d.ts
```diff
@@ -0,0 +1,6 @@
+/// <reference types="next" />
```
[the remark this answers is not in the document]
ketphan02: It's auto-generated by next whenever the project runs, the document says you don't need to care about this!

src/components/MapCard/index.tsx
```diff
@@ -0,0 +1,57 @@
+import * as React from 'react';
+
+import styles from './mapCard.module.scss';
+
+export type MapCardProps = {
+	title: string;
+	desc: string;
+	diff: 'Easy' | 'Medium' | 'Hard';
```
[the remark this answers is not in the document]
ketphan02: Sure, will do!

src/components/MapCard/index.tsx
ketphan02: Ah yeah this is just me afraid of callback hell

ketphan02: I shouldn't put this inside a component

ketphan02:
image
Fixed card

src/components/Global/Card/index.tsx
```diff
@@ -0,0 +1,74 @@
+import * as React from 'react';
+
+import { Button } from '../';
+
+import styles from './mapCard.module.scss';
+
+type HARD = 'hard';
+type MEDIUM = 'medium';
+type EASY = 'easy';
+type DIFFICULTY = EASY | MEDIUM | HARD;
```
Opeyem1a, line 10: My last review comment was meant to mean that I don't think this hassle is worthwhile. We're still doing `x === string literal` later on in the code so defining these as types doesn't really benefit us more than if we just did:
```ts
export type CardProps = {
  title: string;
  desc: string;
  diff: 'easy' | 'medium' | 'hard';
  type?: 'list' | 'grid';
};
```

src/components/Global/Card/mapCard.module.scss
```diff
@@ -0,0 +1,63 @@
...
+  background-color: $white;
+  border-radius: 0.8rem;
+  border: 0.15rem solid $black;
+  width: 28rem;
+  padding: 1.5rem 1.25rem;
+  margin: 1.25rem;
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+  @include font-body;
```
Opeyem1a, line 22: Let's put `@includes` at the beginning of class definitions

Opeyem1a (changes requested):
Minor changes and then I think this is good to merge!

src/components/Global/Card/index.tsx
ketphan02: Yeah but I do this because there are some props in the functions still use the type DIFFICULTY

ketphan02:
A little bit messy since I forgot to run prettier last time

Opeyem1a (approved):
LGTM

ketphan02:
FInally 🙏🏻

PR #17, opened 2021-10-17
Header Component

ketphan02:


ketphan02:
image
Example

src/components/Global/Header/Header.module.scss
```diff
@@ -0,0 +1,76 @@
+$white: white;
+$black: black;
+$primary-color: #F16B67;
+
+.header {
+    width: 100vw;
+    height: 7vh;
+    background-color: $white;
+    padding: 0 2vw;
```
MtheDV, line 9: Header should be a fixed height (rem based). If you shrink the window height the header will shrink which could make it unusable.

src/components/Global/Header/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+                        <span className={ activeTab === 1 && styles.active } onClick={() => setActiveTab(1)}>
+                            <p>Puzzles List</p>
+                        </span>
+                    </div>
+                </div>
+                <div className={ styles.right }>
+                    <button> Profile </button>
+                    {   profilePicture
+                        ? <img  className={ styles.profileImg } src={ profilePicture } alt="profile" />
+                        : <span className={ styles.profileImg }> </span>
```
MtheDV, line 35: would be cool to add text inside the profile image (like the first letter of their username/email) if they don't have one so there is at least something there

MtheDV: But this could just be another feature and doesn't need to be in this one

src/components/Global/Header/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+    const [activeTab, setActiveTab] = React.useState<0 | 1>(0);
+
+    return (
+        <div className={ styles.header }>
+            <div className={ styles.content}>
+                <div className={ styles.left }>
+                    <>
+                        <Image src={ logo } width={57} height={57} /> { /** This is a mock, change when there's logo component  */ }
+                        <p>Computational Puzzles for Kids</p>
+                    </>
```
MtheDV, line 21: Still needs to be changed now that it's implemented

src/components/Global/Header/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+                        ? <img  className={ styles.profileImg } src={ profilePicture } alt="profile" />
+                        : <span className={ styles.profileImg }> </span>
+                    }
+                </div>
+            </div>
+        </div>
+    );
+}
+
+export default Header;
```
MtheDV, line 43: Make sure to run `npm run prettify` to format your code

src/components/Global/Header/Header.module.scss
```diff
@@ -0,0 +1,76 @@
...
+            }
+
+            .tabs {
+                margin-left: 5rem;
+                cursor: pointer;
+                display: flex;
+                align-items: center;
+
+                span {
+                    height: 6vh;
```
MtheDV, line 38: same thing, this should probably be based off of the height of the header instead of a set value, since it's a component inside of the header

src/components/Global/Header/Header.module.scss
```diff
@@ -0,0 +1,76 @@
...
+                    transition: all .5s;
```
MtheDV, line 39: we have a mixin for setting transition

src/components/Global/Header/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+                        <Image src={ logo } width={57} height={57} /> { /** This is a mock, change when there's logo component  */ }
+                        <p>Computational Puzzles for Kids</p>
+                    </>
+                    <div className={ styles.tabs }>
+                        <span className={ activeTab === 0 && styles.active } onClick={() => setActiveTab(0)}>
+                            <p>Puzzles Map</p>
+                        </span>
+                        <span className={ activeTab === 1 && styles.active } onClick={() => setActiveTab(1)}>
+                            <p>Puzzles List</p>
+                        </span>
```
MtheDV, line 28: the `p` tags should be `span` tags and the `span` tags should be `button`s

src/components/Global/Header/Header.module.scss
```diff
@@ -0,0 +1,76 @@
...
+            .profileImg {
+                height: 2rem;
+                width: 2rem;
+                border-radius: 10rem;
+                margin-left: 1rem;
+                border: 0.13rem solid $black;
+            }
+        }
+    }
+}
```
MtheDV, line 76: Eventually (based on the homepage branch which will be added once buttons are done), there will be mixins for setting the styles at certain breakpoints. Which this will need to include as well

src/components/Global/Header/Header.module.scss
```diff
@@ -0,0 +1,76 @@
+$white: white;
+$black: black;
+$primary-color: #F16B67;
```
MtheDV, line 3: There are variables for these now

src/components/Global/Header/Header.module.scss
```diff
@@ -0,0 +1,76 @@
...
+
+.header {
+    width: 100vw;
+    height: 7vh;
+    background-color: $white;
+    padding: 0 2vw;
+
```
MtheDV, line 10: header should also have a small grey border at the bottom. `border-bottom: rgba($grey, 0.05) 1px solid;` is what I use

src/components/Global/Header/Header.module.scss
```diff
@@ -0,0 +1,76 @@
...
+                    p {
+                        font-size: 1.15rem;
+                        font-weight: 550;
+                        margin: 0 0.5rem;
+                    }
+                }
+
+                .active {
+                    border-bottom: 0.35rem solid $primary-color;
+                }
```
MtheDV, line 52: this class should be inside the tabs class since it relates to that and not the left content

MtheDV (changes requested):

Opeyem1a (changes requested):

Opeyem1a (approved):
We can resolve the functionality of this component after merging it in

MtheDV (approved):

PR #39, opened 2021-11-01
Filter component

### LINKED ISSUE #38: Create radio button & puzzle filter

yuqi88:
- Create radio button component
- Add style to the radio button when toggled
- Create puzzle filter component using the radio button component

Closes #38


src/components/Global/Filter/index.tsx
```diff
@@ -0,0 +1,53 @@
+import React, { useState } from 'react';
+import RadioButton from '../RadioButton';
+import styles from './Filter.module.scss';
+
+const Index = () => {
```
ketphan02, line 5: change the component name to Filter

src/components/Global/Filter/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const [easyChecked, setEasyChecked] = useState(false);
+  const [mediumChecked, setMediumChecked] = useState(false);
+  const [hardChecked, setHardChecked] = useState(false);
+
```
ketphan02, line 9: Just a suggestion
How about having an array of 3 and you will update the element inside that array
```js
const [levelChecker, setLevelChecker] = useState<Array<Boolean>>([false, false, false]);
```
and update will be
```js
const temp = [...levelChecker];
temp[1] = true;
setLevelChecker(temp);
```

ketphan02: not sure if it will run faster but it will have less state to manage

src/components/Global/RadioButton/RadioButton.module.scss
```diff
@@ -0,0 +1,64 @@
+.gone {
+  position: absolute;
+  opacity: 0;
+  cursor: pointer;
+  height: 0;
+  width: 0;
+}
+.normal {
+  display: inline-block;
+  padding: 10px 20px;
```
ketphan02, line 10: change px to rem

src/components/Global/RadioButton/RadioButton.module.scss
```diff
@@ -0,0 +1,64 @@
...
+  @extend .normal;
+  border-color: #d7222c;
+  color: #d7222c;
+
+  &:hover {
+    text-decoration: underline;
+  }
+}
+//checked style
+//easy:checked ~ label{
```
ketphan02, line 46: you should remove the comments if it's a draft

src/components/Global/RadioButton/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+  name: string /*filter*/;
+  difficulty: string /*Easy>> */;
+  checked: boolean;
+  setChecked: (wasChecked: any) => void;
+};
+const getLabelColor = (difficulty: string, isChecked: boolean) => {
+  // console.log(difficulty + " "+ isChecked);
+  if (difficulty === 'Easy') {
+    if (isChecked) {
+      // console.log(isChecked + " in rdobtn @1@");
```
ketphan02, line 14: also remove these console.log

src/components/Global/RadioButton/index.tsx
```diff
@@ -0,0 +1,59 @@
+import styles from './RadioButton.module.scss';
+
+type RadioButtonProps = {
+  id: string /*radioEasy*/;
+  name: string /*filter*/;
+  difficulty: string /*Easy>> */;
+  checked: boolean;
+  setChecked: (wasChecked: any) => void;
+};
+const getLabelColor = (difficulty: string, isChecked: boolean) => {
```
ketphan02, line 10: I think you can improve this function by separating checking "checked" and checking difficulty. You can check the difficulty first then check if checked, then merge them later.

src/components/Global/RadioButton/index.tsx
```diff
@@ -0,0 +1,59 @@
+import styles from './RadioButton.module.scss';
+
+type RadioButtonProps = {
+  id: string /*radioEasy*/;
+  name: string /*filter*/;
+  difficulty: string /*Easy>> */;
```
ketphan02, line 6: create a type like
```js
type DIFFICULTY = 'hard' | 'medium' | 'easy';
```
and replace string with DIFFICULTY type

src/components/Global/RadioButton/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+  checked: boolean;
+  setChecked: (wasChecked: any) => void;
```
ketphan02, line 8: change wasChecked type to boolean

styles/login.module.scss
```diff
@@ -0,0 +1,37 @@
...
+}
+.loginPage button{
+  width:  100%;
+  background-color: coral;
+  color: white;
+  border: 0;
+  padding: 8px;
+  border-radius:  8px;
+  cursor: pointer;
+}
```
ketphan02, line 37: add a new line here

src/components/Global/RadioButton/index.tsx
ketphan02: also setChecked is a function from useState if you use it like above, it means that the type of this function is not `() => void`, it's `React.Dispatch<React.SetStateAction<boolean>>`

ketphan02: This function might have both cases, you can handle this by `React.Dispatch<React.SetStateAction<string>> | (wasChecked: boolean) => void;`

ketphan02 (changes requested):

styles/login.module.scss
yuqi88: deleted the file in this branch. 
login.tsx is being reviewed in #39

src/components/Global/Filter/index.tsx
yuqi88: it's recommended to separate the states, so I'll leave them the way they are.
https://reactjs.org/docs/hooks-faq.html#should-i-use-one-or-many-state-variables

src/components/Global/RadioButton/RadioButton.module.scss
```diff
@@ -0,0 +1,63 @@
...
+  height: 0;
+  width: 0;
+}
+.normal {
+  display: inline-block;
+  padding: 0.625rem 1.25rem;
+  margin: 0.625rem 1.25rem;
+  font-family: 'Nunito Sans', serif;
+  //font-size: 1rem;
+  border: 2px solid;
```
ketphan02, line 14: change to rem

src/components/Global/RadioButton/RadioButton.module.scss
```diff
@@ -0,0 +1,63 @@
...
+.mediumChecked {
+  @extend .medium;
+  text-decoration: initial;
+  background-color: #3bb2c8;
+  color: white;
+}
+.hardChecked {
+  @extend .hard;
+  text-decoration: initial;
+  background-color: #d7222c;
```
ketphan02, line 61: import the colors from our mixin style

src/pages/index.tsx
```diff
@@ -16,6 +17,7 @@ const Home: React.FC = () => {
     <>
       <h1>Not logged in</h1>
       <button onClick={() => signIn()}>Login</button>
+      <Filter />
```
ketphan02, line 20: I think you should remove this

src/components/Global/RadioButton/RadioButton.module.scss
```diff
@@ -0,0 +1,63 @@
...
+  text-align: center;
+  cursor: pointer;
+}
+.easy {
+  @extend .normal;
+  border-color: #05c05b;
+  color: #05c05b;
+
+  &:hover {
+    text-decoration: underline;
```
ketphan02, line 25: this is the same for all the button so I think you can move this to the .normal

ketphan02 (changes requested):

src/components/Global/Filter/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+        />
+        <RadioButton
+          id={'radioHard'}
+          name={'filter'}
+          difficulty={'hard'}
+          checked={hardChecked}
+          setChecked={setHardChecked}
+        />
+      </form>
+      <button
```
Opeyem1a, line 35: This button should be a real `<Button>` that says "Search" or something

src/components/Global/Filter/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+import styles from './Filter.module.scss';
+
+const Filter = () => {
+  const [easyChecked, setEasyChecked] = useState(false);
+  const [mediumChecked, setMediumChecked] = useState(false);
+  const [hardChecked, setHardChecked] = useState(false);
+
+  return (
+    <div>
+      <form className={styles.filter}>
```
Opeyem1a, line 12: The `Filter` component should also contain an `<Input>` and the state of that input method should also be managed here. You'll probably need to merge the updated version (after resolving reviews) of the Input component into this branch again

src/components/Global/RadioButton/RadioButton.module.scss
```diff
@@ -0,0 +1,57 @@
...
+  border-color: $secondary-100;
+  color: $secondary-100;
+}
+.hard {
+  @extend .normal;
+  border-color: $red;
+  color: $red;
+}
+.easyChecked {
+  @extend .easy;
```
Opeyem1a, line 41: `@extend` here isn't necessary, you could just add or remove the `easyChecked` class and leave the `easy` class on the component. The current way you have means the `easy` styles are duplicated when an element has both `easy` and `easyChecked`

This applies to `mediumChecked` and `hardChecked`

src/components/Global/RadioButton/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+import styles from './RadioButton.module.scss';
+
+type DIFFICULTY = 'easy' | 'medium' | 'hard';
+
+type RadioButtonProps = {
+  id: string /*radioEasy*/;
+  name: string /*filter*/;
+  difficulty: DIFFICULTY;
+  checked: boolean;
+  setChecked: React.Dispatch<React.SetStateAction<boolean>>;
```
Opeyem1a, line 11: Oh is this the proper way to do this? It looks kinda clunky

src/components/Global/RadioButton/index.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import styles from './RadioButton.module.scss';
+
+type DIFFICULTY = 'easy' | 'medium' | 'hard';
+
+type RadioButtonProps = {
+  id: string /*radioEasy*/;
```
Opeyem1a, line 7: the comment here and on line 8 should be removed

src/components/Global/RadioButton/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+  name: string /*filter*/;
+  difficulty: DIFFICULTY;
+  checked: boolean;
+  setChecked: React.Dispatch<React.SetStateAction<boolean>>;
+};
+const getLabelColor = (difficulty: DIFFICULTY, isChecked: boolean) => {
+  return isChecked ? styles[`${difficulty}Checked`] : styles[`${difficulty}`];
```
Opeyem1a, line 14: Even though this is a beautiful 1-line code, it creates a dependency between the name of the `DIFFICULTY` types and the names of the classes that I don't think we should have.

So I think the move is:
```ts
if(difficulty === 'easy') return isChecked ? styles.easyChecked : styles.easy;
// repeat for 'medium' and 'hard'
```

src/components/Global/RadioButton/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+  return (
+    <div className="radio">
+      <input
+        className={styles.gone}
+        type="radio"
+        id={id}
+        name={name}
+        value={difficulty}
+        onClick={() => {
+          setChecked(!checked);
```
Opeyem1a, line 33: wait shouldn't it be `setChecked((wasChecked) => !wasChecked);`?

src/pages/index.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import { signIn, signOut, useSession } from 'next-auth/react';
 
-const Home: React.FC = () => {
```
Opeyem1a, line 4: Undo this change since it has nothing to do with this branch and will cause conflicts with other branches

Opeyem1a (changes requested):

src/components/Global/RadioButton/index.tsx
yuqi88: yee. The old one was wrong and more clunky.

src/components/Global/RadioButton/index.tsx
yuqi88: Is dependency sth we want to reduce? If yes, why is that?
In this case, if we use if-statements, the value of label colour also depends on the value of difficulty, so what's the difference?

src/components/Global/Filter/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+        <RadioButton
+          id={'radioEasy'}
+          name={'filter'}
+          difficulty={'easy'}
+          checked={easyChecked}
+          onClick={() => {
+            setEasyChecked(wasChecked => !wasChecked);
+            setFilterFields({
```
Opeyem1a, line 23: This is not how we want to do this, ideally we need to set up a `useEffect` hook that listens for changes to `easyChecked`, `mediumChecked`, and `hardChecked` and then uses `setFilterFields()` to match their values. Look into `useEffect()` for implementation details

src/components/Global/Filter/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+            setHardChecked(wasChecked => !wasChecked);
+            setFilterFields({
+              easy: easyChecked,
+              medium: mediumChecked,
+              hard: !hardChecked
+            });
+          }}
+        />
+      </form>
+      {/*{console.log("e "+ easyChecked+ " / m"+ mediumChecked+" /h"+hardChecked)}*/}
```
Opeyem1a, line 59: Remove unneeded comments

src/components/Global/Input/index.tsx
```diff
@@ -2,20 +2,24 @@ import React, { useState } from 'react';
...
 type InputProps = {
-  type: 'text' | 'password';
+  type: 'text' | 'password' | 'email';
   id: string;
   required: boolean;
   placeholder?: string;
   maxLength?: number;
+  minLength?: number;
   labelText?: string;
+  setInputVal: React.Dispatch<React.SetStateAction<string>>;
```
Opeyem1a, line 12: Is this type necessary? Does `() => void` not work?

src/components/Global/Input/index.tsx
```diff
@@ -28,8 +32,12 @@ const Input = ({
         required={required}
         placeholder={placeholder}
         maxLength={maxLength}
+        minLength={minLength}
         value={input}
-        onChange={event => setInput(event.target.value)}
+        onChange={event => {
+          setInput(event.target.value);
```
Opeyem1a, line 38: If you're using the `setInputVal` technique, this component doesn't need to have an internal `input` state variable

Opeyem1a (changes requested):

Opeyem1a:
Sorry I didn't review this earlier @yuqi88, I didn't notice there were more changes - also resolve the merge conflict on this branch so we can hopefully merge it soon

src/components/Global/Input/index.tsx
yuqi88: - The above way of defining the type is recommended for passing useState function as props though ( ) => void isn't entirely wrong. 
- setInputVal could either take in ( value: string | ( preVal: string) => { } ), and the second type in the return type is a dispatch. 
- Dispatch<SetStateAction<string>> is more generic; thus recommended.

src/components/Global/Input/index.tsx
yuqi88: The value attribute of the input tag is assigned by the value in the input variable. (line before onChange = ....)
the internal input state is still necessary.

src/components/Global/Filter/Filter.module.scss
```diff
@@ -0,0 +1,5 @@
+.filter {
+  display: flex;
+  justify-content: space-between;
+  width: 19rem;
```
Opeyem1a, line 4: why does this have a width at all?

src/pages/index.tsx
Opeyem1a: Reminder to do this

Opeyem1a (changes requested):

Opeyem1a (approved):

PR #40, opened 2021-11-03
Puzzle page

Opeyem1a:


Opeyem1a:
This branch contains schema changes that we need to iron out before merging

src/utils/puzzles.ts
```diff
@@ -0,0 +1,52 @@
...
+  axios.post('/api/puzzles/instances/submit', puzzleSubmissionDetails).then(
+    response => {
+      console.log(response.status);
+      alert(response.data.message);
+    },
+    error => {
+      // TODO: proper alerts
+      console.log(error);
+      console.log(error.status);
+      alert('Oh no - errors.');
```
MtheDV, line 37: oh no, our app, it's broken

MtheDV (approved):
I'm going to assume that this all works as it should, I did look through it but since there's a lot I of course have a hard time finding things. I think we should also just get these PRs merged so that the pieces can be put together

PR #52, opened 2021-11-19
Add admin page with automatic QR

ketphan02:


src/components/App/QRGenerator/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+    scale: 4,
+    width: 4,
+    color: {
+      dark: '#000000ff',
+      light: '#ffffffff'
+    }
+  };
+
+  const [qrRef] = useQRCode({
+    text: text,
```
Opeyem1a, line 17: Should this be called text? Or would something like `link` be better? (applies to ever `text` variable in the file)

src/components/App/QRGenerator/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+  });
+
+  return <canvas ref={qrRef} />;
+};
+
+const QRGenerator = () => {
+  const [text, setText] = React.useState('');
+  return (
+    <>
+      <input
```
Opeyem1a, line 28: Leave yourself a TODO to make this like the Input component

src/components/App/QRGenerator/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+        style={{ width: 100, height: 20 }}
+        type="text"
+        onChange={e => setText(e.currentTarget.value)}
+        value={text}
+        placeholder="Information"
+      />
+      <br />
```
Opeyem1a, line 35: Let's not do spacing like this, I recommend just adding a class to the above element and increasing the bottom margin on it

src/components/App/QRGenerator/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+const QRGenerator = () => {
+  const [text, setText] = React.useState('');
+  return (
+    <>
+      <input
+        style={{ width: 100, height: 20 }}
+        type="text"
+        onChange={e => setText(e.currentTarget.value)}
+        value={text}
+        placeholder="Information"
```
Opeyem1a, line 33: Maybe `"Enter link here..."` or something more user friendly

src/components/App/QRGenerator/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  return <canvas ref={qrRef} />;
+};
+
+const QRGenerator = () => {
+  const [text, setText] = React.useState('');
+  return (
+    <>
+      <input
+        style={{ width: 100, height: 20 }}
```
Opeyem1a, line 29: We should avoid inline styling imo, make classes when you need styles - in the interest of getting this merged, the styling doesn't need to look perfect in order to be merged, but it should be a separate `.scss` file for sure

src/pages/admin/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+      Router.push('/auth/login');
+    }
+  }, [status]);
+
+  if (status === 'loading') {
+    return <div>Loading...</div>;
+  }
+
+  const email = session?.user?.email;
+  if (status === 'authenticated' && email.split('@')[1] === 'ubc.ca') {
```
Opeyem1a, line 20: Why does the email have to end with "ubc.ca"? And if it does, it should be a constant defined globally elsewhere.
Like we should have permission functions like `isAdminUser()` that return booleans and the logic of defining an admin user should be there

src/pages/admin/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const { data: session, status } = useSession();
+
+  React.useEffect(() => {
+    if (status === 'unauthenticated') {
+      Router.push('/auth/login');
+    }
+  }, [status]);
+
+  if (status === 'loading') {
+    return <div>Loading...</div>;
```
Opeyem1a, line 16: I didn't know we had a loading status, this is cool!

Opeyem1a (changes requested):

src/components/App/QRGenerator/index.tsx
ketphan02: Yeah because it can basically be anything, I think we need a discussion for whether we want it to be a link or not, then I'll add a regexp to filter links.

src/components/App/QRGenerator/index.tsx
ketphan02: My bad, just didn't really want to do CSS 😅

MtheDV (approved):
Looking through it looks good!

ketphan02:
Aight, let's wait for @Opeyem1a to take another look then

src/pages/admin/index.tsx
```diff
@@ -0,0 +1,50 @@
+import * as React from 'react';
+import Router from 'next/router';
+import { useSession } from 'next-auth/react';
+import { QRGenerator } from '../../components/App';
+import { isAdmin } from '../../services/admin';
+
+const Admin = () => {
```
Opeyem1a, line 7: `AdminPage`

src/pages/api/admin/validate.ts
```diff
@@ -0,0 +1,33 @@
...
+        message: 'User not found'
+      });
+    }
+    const account = await prisma.account.findMany({
+      where: {
+        userId: user.id,
+        type: 'admin'
+      }
+    });
+    if (account.length > 0) {
```
Opeyem1a, line 23: What if multiple accounts are returned - we should throw an error in this case

src/pages/admin/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+        <QRGenerator />
+      </>
+    ) : (
+      <h1> You are not allowed here </h1>
+    ); // This will need adjustment since it's just a prototype
+  }
+
+  return (
+    <>
+      {/** TODO: Create Header for admin page  */}
```
Opeyem1a, line 43: Do we need this, the code seems to redirect unauthenticated users back to the login page anyway.

It's worth having a version of the 403 component that handles what this displays

src/pages/admin/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  if (status === 'authenticated') {
+    return validAdmin ? (
+      <>
+        <h1> ADMIN PAGE 🤓 </h1>
+        {/** TODO: Create Header for admin page  */}
+        <QRGenerator />
+      </>
+    ) : (
+      <h1> You are not allowed here </h1>
```
Opeyem1a, line 37: 403 component

src/services/admin.ts
```diff
@@ -0,0 +1,13 @@
+import Axios from './axios';
+
+export const isAdmin = async (email: string) => {
```
Opeyem1a, line 3: Do we need types on functions like this?

Opeyem1a (changes requested):

src/pages/admin/index.tsx
Opeyem1a: Oh wait I'm bad we don't name them like this currently

src/pages/admin/index.tsx
MtheDV: Yeah we should just 403 if they're not authorized

src/services/admin.ts
ketphan02: We'll talk later, but i think the backend is the one who need to declare the type at the first place.

src/components/Global/ErrorDisplay/index.tsx
```diff
@@ -12,7 +13,7 @@ const ErrorDisplay = ({
   return (
     <div className={styles.center}>
       <div className={styles.contentWrapper}>
-        <div className={styles.logo}>
+        <div className={styles.logo} onClick={ () => Router.push('/') } >
           <Logo showMark={true} showType={true} />
```
Opeyem1a, line 17: The new Logo component in main has a prop you can pass to make it clickable and returns to home

Opeyem1a (changes requested):

Opeyem1a (approved):

PR #53, opened 2021-11-19
Add 404 page

ketphan02:


Opeyem1a (changes requested):
It generally looks solid to me for now (the error display might need to have more props in the future for redirects, extra info and whatnot) -- Could you add some screenshots to the PR so we can do a quick visual pass on it before merging?

ketphan02:
Sure
image

ketphan02:
Yeah I was thinking that too, I think we need a button to redirect to the landing page

MtheDV (changes requested):
Based on the screenshot: I think we could have the app image at the top, then underneath it show the 404 error with the quote. Then later we can add stuff, but for now that would be good.

ketphan02:
image
This looks better than the last time ngl

Opeyem1a (approved):

PR #64, opened 2021-12-02
Map Renderer Functionality

Opeyem1a:
This is ready and should be the first one to be merged. It's getting too big and needs to be dealt with at this point.


MtheDV:
Need to wait for filter component to be added

PR #73, opened 2021-12-09
Generate puzzle instance

ketphan02:


ketphan02:
@Opeyem1a can you help me out with creating the puzzle instance API, I'm not sure how to do that

Opeyem1a:
Sure I gotchu

prisma/schema.prisma
```diff
@@ -128,7 +128,7 @@ model Submission {
   isCorrect        Boolean[]      @map(name: "is_correct")
   randomSeed       Float[]        @map(name: "random_seed")
 
-  @@unique([puzzleInstanceId, userId])
+  @@unique([puzzleInstanceId, userId], map: "puzzleInstanceId_userId")
```
Opeyem1a, line 131: Do we need this? I thought this would be the default map name

src/components/App/PuzzleGenerate/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+    const puzzle = event.target[4].value;
+
+    console.log(hint, latitude, longitude, address, puzzle);
+    createPuzzleInstance(puzzle, longitude, latitude, address, hint).then(r => console.log(r));
+  };
+
+  const [puzzleList, setPuzzleList] = useState([]);
+
+  useEffect(() => {
+    const fetchPuzzles = async () => {
```
Opeyem1a, line 24: I've seen this pattern somewhere before but why do we do it like this again? Would it be bad to define `fetchPuzzles()` outside of the hook's body and then just call the function within the hook's body?

src/components/App/PuzzleGenerate/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+      setPuzzleList(await getAllPuzzles());
+    };
+    fetchPuzzles();
+  }, []);
+
+  return (
+    <>
+      <form className={styles.form} onSubmit={event => handleSubmit(event)}>
```
Opeyem1a, line 32: I prefer the pattern with state variables and then a function that's called on submit. Doing it this way and then using arrays feels iffy

src/components/App/PuzzleGenerate/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+        <h2> Make a puzzle instance </h2>
+        <input type="text" id="puzzleHint" placeholder="Hint" />
+        <div className={styles.puzzleLocation}>
+          <input type="text" id="puzzleLatitude" placeholder="Latitude" />
+          <input type="text" id="puzzleLongitude" placeholder="Longitude" />
+          <input type="text" id="puzzleAddress" placeholder="Address" />
+        </div>
+        <div>
+          {puzzleList.length > 0 && (
```
Opeyem1a, line 41: I ran into an issue when I ran this - what if there aren't any puzzles?

An idea:
The `<select>` should show up regardless, but since there are no puzzles they won't be able to select a value, and if we make this field required then they won't be able to submit the form at all if there are no puzzles

Opeyem1a: Also couldn't we do this in the falsey way like `puzzleList && ...`? I think I remember reading somewhere that `.length > 0` is better but I forgot why -- look into this

src/components/App/PuzzleGenerate/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+                  <option value={puzzle.id} key={`puzzle${index}`}>
+                    {puzzle.name}
+                  </option>
+                ))}
+              </optgroup>
+            </select>
+          )}
+        </div>
+        <div>
+          {/* <input type='submit' value='Submit' /> */}
```
Opeyem1a, line 54: Remove unneeded comments

Opeyem1a (changes requested):

src/components/App/PuzzleGenerate/index.tsx
Opeyem1a: We should use the Input component when we switch it over, helps with consistent styling, too

prisma/schema.prisma
ketphan02: it got error when I check this function, so I added this ( I think eslint did that)

src/components/App/PuzzleGenerate/index.tsx
ketphan02: I'm not sure about this but I think if you put this function inside of useEffect, it will not rerender when the component rerender.

src/components/App/PuzzleGenerate/index.tsx
ketphan02: how did this happened, I tried fasley way will not work since [] is not faley and I always return [] if the API did not work

ketphan02: I deleted my puzzle in the database but then it disappear right away hmmmmm 🤔

src/components/App/PuzzleGenerate/index.tsx
Opeyem1a: I'll look into this later

src/components/App/PuzzleGenerate/index.tsx
Opeyem1a: Deleted your puzzle?

src/components/App/PuzzleGenerate/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+          <Input type='text' id='puzzleLatitude' required={ true } placeholder='Latitude' setInputVal={ setLatitude } />
+          <Input type='text' id='puzzleLongitude' required={ true } placeholder='Longitude' setInputVal={ setLongitude } />
+          <Input type='text' id='puzzleAddress' required={ true } placeholder='Address' setInputVal={ setAddress } />
+        </div>
+        <div>
+          { puzzleList.length > 0 && (
+            <select className={ styles.selections } value={ puzzle } onChange={ (e) => setPuzzle(e.currentTarget.value) }>
+              <optgroup label="Choose a puzzle from the puzzles below">
+                { puzzleList.map((pzl, index) => (
+                  <option value={ pzl.name } key={ `puzzle${index}` }>
```
Opeyem1a, line 44: puzzle names are not unique, and the API searches for puzzles by id. So the value should be the puzzle's id

Opeyem1a (changes requested):

src/components/App/PuzzleGenerate/index.tsx
ketphan02: sure, but I think as a user, I won't understand what's going there

ketphan02: What do you think @Opeyem1a 🤔

src/components/App/PuzzleGenerate/index.tsx
ketphan02: The one I created manually

src/components/App/PuzzleGenerate/index.tsx
Opeyem1a: No like we can make `value=pzl.id` but still do `<option ...>{pzl.name}</option>` so the user sees the name

src/components/App/PuzzleGenerate/index.tsx
Opeyem1a: I'm not sure what you mean but the original comment is already resolved

src/components/App/PuzzleGenerate/index.tsx
ketphan02: Oh yeah got it!

src/components/App/PuzzleGenerate/index.tsx
```diff
@@ -0,0 +1,101 @@
+import * as React from 'react';
+import { useEffect, useState } from 'react';
+import styles from './PuzzleGenerate.module.scss';
+import { Button, Input } from '../../Global';
+import { createPuzzleInstance, getAllPuzzles } from '../../../services';
+
+const PuzzleGenerate = () => {
+  const [puzzleList, setPuzzleList] = useState([]);
+  const [hint, setHint] = useState('');
+  const [latitude, setLatitude] = useState('');
```
Opeyem1a, line 10: why are we tracking latitude, longitude, and id as strings in the state? Aren't they numbers?

src/components/App/PuzzleGenerate/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const [longitude, setLongitude] = useState('');
+  const [address, setAddress] = useState('');
+  const [puzzleId, setPuzzleId] = useState('');
+
+  const handleSubmit = () => {
+    console.log(hint, latitude, longitude, address, puzzleId);
+    createPuzzleInstance(
+      parseInt(puzzleId),
+      parseInt(longitude),
```
Opeyem1a, line 19: latitude and longitude are doubles, not integers

src/components/App/PuzzleGenerate/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+      address,
+      hint
+    ).then(puzzleInstance => console.log(puzzleInstance));
+  };
+
+  useEffect(() => {
+    const fetchPuzzles = async () => {
+      const puzzles = await getAllPuzzles();
+      setPuzzleList(puzzles);
+      setPuzzleId(puzzles[0].id);
```
Opeyem1a, line 30: Why?

Opeyem1a: Also what if the list is empty?

Opeyem1a (changes requested):

src/components/App/PuzzleGenerate/index.tsx
MtheDV: floats*

src/components/App/PuzzleGenerate/index.tsx
MtheDV: they are numbers

src/components/App/PuzzleGenerate/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+    };
+    fetchPuzzles();
```
MtheDV, line 32: this doesn't need to be a function if you're going to only call it in this effect and immediately after

src/components/App/PuzzleGenerate/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+const PuzzleGenerate = () => {
+  const [puzzleList, setPuzzleList] = useState([]);
+  const [hint, setHint] = useState('');
+  const [latitude, setLatitude] = useState('');
+  const [longitude, setLongitude] = useState('');
+  const [address, setAddress] = useState('');
+  const [puzzleId, setPuzzleId] = useState('');
+
+  const handleSubmit = () => {
+    console.log(hint, latitude, longitude, address, puzzleId);
```
MtheDV, line 16: remove log

src/components/App/PuzzleGenerate/index.tsx
ketphan02: I needed to resolve the promise of getAllPuzzles() using await

src/components/App/PuzzleGenerate/index.tsx
ketphan02: This log is a placeholder for another feature!

src/components/App/PuzzleGenerate/index.tsx
ketphan02: The options does not give me the default value, so I do this manually

ketphan02: And yes I forgot to handle empty :(

src/components/App/PuzzleGenerate/index.tsx
Opeyem1a: Could't this be done using `.then()` instead?

src/components/App/PuzzleGenerate/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+        address,
+        hint
+      ).then(puzzleInstance => console.log(puzzleInstance));
+  };
+
+  useEffect(() => {
+    const fetchPuzzles = async () => {
+      const puzzles = await getAllPuzzles();
+      setPuzzleList(puzzles);
+      puzzles.length > 0 && setPuzzleId(puzzles[0].id);
```
Opeyem1a, line 31: But like, what happens if you just don't do the setPuzzleId part here? This feels bad because it means any time this useEffect is triggered, it will default the selection back to the first puzzle in the list, even if the user had something else selected, right?

Why do we need to set the puzzle id value here at all?

Opeyem1a (changes requested):

src/components/App/PuzzleGenerate/index.tsx
ketphan02: This useEffect will only be triggered when you first load the page, don't worry. So initially, the puzzleId was null, but I designed the <option> to select the first puzzle as a default. This is like setting useState('') for textArea.

src/components/App/PuzzleGenerate/index.tsx
ketphan02: it could but I just want to use async await, should I change?

src/components/App/PuzzleGenerate/index.tsx
Opeyem1a: Oh wait but then let's just do the call in `getServerSideProps()`. There's no need for a useEffect hook that's only called once when the page loads.

Just pass a list of puzzles as props to this page

ketphan02: Woah I did not even know that exists 🤩

MtheDV (approved):

Opeyem1a (approved):

PR #77, opened 2021-12-14
Puzzle page feedback

Opeyem1a:


src/components/App/FeedbackGif/FeedbackGif.module.scss
```diff
@@ -0,0 +1,23 @@
...
+
+.gifEmbed {
+  border-left: 0.6rem solid;
+  border-right: 0.6rem solid;
+  border-radius: 0.2rem;
+  height: auto;
+  width: auto;
+
+  div {
+    img {
```
ketphan02, line 11: is this img the same as Image of next/image

src/components/Global/LogStatusButton/index.tsx
```diff
@@ -0,0 +1,13 @@
+import { Button } from '../index';
+import { signIn, signOut } from 'next-auth/react';
+import React from 'react';
+
+const LogStatusButton = ({ status }) => {
+  return status === 'authenticated' ? (
+    <Button style={'flat'} content={'Logout'} onClick={() => signOut()} />
+  ) : (
+    <Button style={'flat'} content={'Login'} onClick={() => signIn()} />
```
ketphan02, line 9: I think we should put a TODO: redirect to /auth/login for all functions like this

src/pages/auth/login/index.tsx
```diff
@@ -26,7 +28,8 @@ export default function LoginPage({ providers, csrfToken }) {
       <div className={styles.mainSec}>
         <h2 className={styles.title}>Login</h2>
         {error && <SignInError error={error} />}
-        <form method="post" action="/api/auth/callback/credentials"> {/*TODO: save session into db*/}
+        <form method="post" action={'/api/auth/callback/credentials'}>
```
ketphan02, line 31: i think we should add a login service here

ketphan02 (changes requested):

src/components/App/FeedbackGif/FeedbackGif.module.scss
Opeyem1a: yep

src/pages/auth/login/index.tsx
Opeyem1a: What do you mean?

src/components/Global/LogStatusButton/index.tsx
Opeyem1a: Like redirect instead of the `signIn()`/`signOut()`?

MtheDV (approved):

PR #79, opened 2021-12-14
Search and filter component

### LINKED ISSUE #78: Create search and filter component

yuqi88:
- Add search-and-filter component.
- merge this component before the puzzle list page

closes #78

### DIFF
--- a/src/components/Global/SearchAndFilter/SearchAndFilter.module.scss
+++ b/src/components/Global/SearchAndFilter/SearchAndFilter.module.scss
@@ -1,8 +1,8 @@
 .container {
   display: flex;
   justify-content: space-between;
+  gap: 0.5rem;
 }
-
 .input {
-  width: 26.75rem;
+  flex: 1;
 }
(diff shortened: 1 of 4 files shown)


src/components/Global/SearchAndFilter/index.tsx
```diff
@@ -0,0 +1,46 @@
+import React, { useEffect, useState } from 'react';
+import { Filter, Input } from '../index';
+import styles from './SearchAndFilter.module.scss';
+
+type SearchAndFilterProps = {
+  setSearchNFilterVal?: React.Dispatch<React.SetStateAction<object>>;
```
Opeyem1a, line 6: Do the thing where we import differently to avoid the `React.` twice here

src/components/Global/SearchAndFilter/SearchAndFilter.module.scss
```diff
@@ -0,0 +1,8 @@
+.container {
+  display: flex;
+  justify-content: space-between;
+}
+
+.input {
+  width: 26.75rem;
```
Opeyem1a, line 7: Do we need to manually fix this width? It's usually bad for responsiveness to do this

Opeyem1a (changes requested):

Opeyem1a (approved):

PR #90, opened 2022-01-12
Install + implement SWR

Opeyem1a:
+ added Loading component
+ installed and implemented a swr pattern in "/puzzles" page
+ added custom usePuzzleInstances hook
+ removed need for getServerSideProps


src/utils/errorHandler.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import Loading from '../components/Global/Loading';
+import ErrorDisplay from '../components/Global/ErrorDisplay';
```
ketphan02, line 3: `import { Loading, ErrorDisplay } from '../components/Global';

src/utils/swr.ts
```diff
@@ -0,0 +1,24 @@
...
+    const res = await fetch(url)
+    // If the status code is not in the range 200-299, we still try to parse and throw it.
+    if (!res.ok) {
+        const body = await res.json();
+
+        let error;
+        if (body?.message) error = new Error(body.message);
+        else error = new Error('An error occurred while fetching the data.')
+        // Attach extra info to the error object.
+        error['info'] = body;
```
ketphan02, line 15: I think we only need `info` or `message`

package.json
```diff
@@ -45,6 +46,7 @@
     "prettier": "^2.4.1",
     "react-test-renderer": "^17.0.2",
     "ts-jest": "^27.0.6",
-    "typescript": "^4.4.4"
+    "typescript": "^4.4.4",
+    "swr": "^1.1.2"
```
ketphan02, line 50: there's duplicate `swr` here

ketphan02 (changes requested):
Run `npm run prettify`

> Just a question: should we start replacing files in `/services` by SWR, SWR seems cool!

src/utils/swr.ts
Opeyem1a: I think info should contain the body in case there's extra info attached

package.json
Opeyem1a: Thanks 👌🏾

src/utils/swr.ts
Opeyem1a: What do you think about ^? @ketphan02

PR #98, opened 2022-01-14
Add Toaster and fix minor bugs

### LINKED ISSUE #95: Replace alert() with Toaster

ketphan02:
+ Close #95 
+ Remove all alerts with [`react-hot-toaster`](https://react-hot-toast.com)
+ Fix minor bugs about optional
+ Pretify the code

### DIFF
--- a/src/components/App/PuzzleGenerate/index.tsx
+++ b/src/components/App/PuzzleGenerate/index.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { useState } from 'react';
 import styles from './PuzzleGenerate.module.scss';
-import toast, { Toaster } from 'react-hot-toast';
+import toast from 'react-hot-toast';
 
 import { Button, Input } from '../../Global';
 import { QRGenerator } from '..';
@@ -32,7 +32,7 @@ const PuzzleGenerate = ({ puzzlesList }) => {
       }
     };
     toast.promise(puzzleInstancePromise(), {
-      loading: 'Hello',
+      loading: 'Making your puzzle instance... ⚙️',
       success: 'Success',
       error: 'Something went wrong :('
     });
@@ -97,7 +97,6 @@ const PuzzleGenerate = ({ puzzlesList }) => {
             arrowDirection="right"
             onClick={() => handleSubmit()}
           />
-          <Toaster />
         </div>
         {puzzleInstanceData && (
           <div className={styles.qrCode}>
--- a/src/components/App/PuzzleGenerate/index.tsx
+++ b/src/components/App/PuzzleGenerate/index.tsx
@@ -31,84 +31,81 @@ const PuzzleGenerate = ({ puzzlesList }) => {
         setPuzzleInstanceData(puzzleInstance);
       }
     };
-    toast.promise(
-      puzzleInstancePromise(),
-      {
-        loading: 'Hello',
-        success: 'Success',
-        error: 'Something went wrong :(',
-      }
-    );
+    toast.promise(puzzleInstancePromise(), {
+      loading: 'Hello',
+      success: 'Success',
+      error: 'Something went wrong :('
+    });
   };
 
   return (
     <>
-      <div className={ styles.form }>
+      <div className={styles.form}>
         <h2> Make a puzzle instance </h2>
         <Input
           type="text"
           id="puzzleHint"
-          required={ true }
+          required={true}
           placeholder="Hint"
-          setInputVal={ setHint }
+          setInputVal={setHint}
         />
-        <div className={ styles.puzzleLocation }>
+        <div className={styles.puzzleLocation}>
           <Input
             type="text"
             id="puzzleLatitude"
-            required={ true }
+            required={true}
             placeholder="Latitude"
-            setInputVal={ setLatitude }
+            setInputVal={setLatitude}
           />
           <Input
             type="text"
             id="puzzleLongitude"
-            required={ true }
+            required={true}
             placeholder="Longitude"
-            setInputVal={ setLongitude }
+            setInputVal={setLongitude}
           />
           <Input
             type="text"
             id="puzzleAddress"
-            required={ true }
+            required={true}
             placeholder="Address"
-            setInputVal={ setAddress }
+            setInputVal={setAddress}
           />
         </div>
         <div>
-          { puzzlesList.length > 0 && (
+          {puzzlesList.length > 0 && (
             <select
-              className={ styles.selections }
-              value={ puzzleId }
-              onChange={ e => setPuzzleId(e.currentTarget.value) }
+              className={styles.selections}
+              value={puzzleId}
+              onChange={e => setPuzzleId(e.currentTarget.value)}
             >
               <option selected>Choose a puzzle</option>
-              { puzzlesList.map((puzzle, index) => (
-                <option value={ puzzle.id } key={ `puzzle${index}` }>
-                  { puzzle.name }
+              {puzzlesList.map((puzzle, index) => (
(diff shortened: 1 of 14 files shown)


src/components/App/PuzzleGenerate/index.tsx
```diff
@@ -1,31 +1,41 @@
...
+      );
+
+      if (puzzleInstance.error) {
+        throw new Error(puzzleInstance.message);
+      } else {
+        setPuzzleInstanceData(puzzleInstance);
+      }
+    };
+    toast.promise(puzzleInstancePromise(), {
+      loading: 'Hello',
```
Opeyem1a, line 35: Hello?

src/components/App/PuzzleGenerate/index.tsx
```diff
@@ -87,6 +97,7 @@ const PuzzleGenerate = ({ puzzlesList }) => {
             arrowDirection="right"
             onClick={() => handleSubmit()}
           />
+          <Toaster />
```
Opeyem1a, line 100: Don't we put this in _document or _page?

src/pages/auth/signup/index.tsx
```diff
@@ -13,17 +14,48 @@ const SignUpPage = () => {
...
+        }
       );
-      event.preventDefault();
       return false;
     }
     if (status === 'authenticated') {
-      if (confirm('Would you like to make another account?'))
-        alert('Please log out before you make an other account.');
-      event.preventDefault();
+      toast(t => (
```
Opeyem1a, line 29: Are you sure this implementation replaces the `confirm()` call above?

src/pages/puzzles/[instanceId]/index.tsx
```diff
@@ -38,42 +41,63 @@ const PuzzlePage = ({
...
+            throw new Error('Almost there! Try Again');
+          }
+        }
+      };
 
-      setIsRecentCorrect(success);
-      setFeedbackGifSrc(getRandomGifSrc(feedbackGifs, success));
+      toast.promise(submissionInstance(), {
+        loading: 'Submitting your answer... ⏳',
+        success: 'Correct! Nice work!',
```
Comment (Opeyem1a) [inline src/pages/puzzles/[instanceId]/index.tsx:74]: It seems weird to throw an error for incorrect answers - isn't there a way to grab whatever the promise returns and check `isCorrect`?

src/pages/puzzles/[instanceId]/index.tsx
```diff
@@ -38,42 +41,63 @@ const PuzzlePage = ({
...
+        </span>
+      ));
       // TODO: should be an alert dialog w/ link to login page
     }
   };
 
   return (
     <>
       <Header />
+      <Toaster />
```
Comment (Opeyem1a) [inline src/pages/puzzles/[instanceId]/index.tsx:100]: Yeah I don't think we need to put this tag in multiple places

Opeyem1a (changes requested):

src/components/App/PuzzleGenerate/index.tsx
ketphan02: Oops 🥲

src/pages/puzzles/[instanceId]/index.tsx
Comment (ketphan02) [reply src/pages/puzzles/[instanceId]/index.tsx:74]: I did try to use the hooks approach but somehow toast does not get the update from `useState`.
Normally, I will have `const [errMsg, setErrMsg] = useState<string>();` and then change it according to the cases. But for toast, it doesn't get the update from `setErrMsg` and will fail to send the message. So I think of this solution.
If you think of a better way, please tell me, I know this is an awful implementation 🥺

src/components/App/PuzzleGenerate/index.tsx
```diff
@@ -1,31 +1,41 @@
...
+      if (puzzleInstance.error) {
+        throw new Error(puzzleInstance.message);
+      } else {
+        setPuzzleInstanceData(puzzleInstance);
+      }
+    };
+    toast.promise(puzzleInstancePromise(), {
+      loading: 'Making your puzzle instance... ⚙️',
+      success: 'Success',
+      error: 'Something went wrong :('
```
Opeyem1a, line 37: Is there at least a way to display the actual message from the error and not just the default string here?

src/components/App/PuzzleGenerate/index.tsx
```diff
@@ -1,31 +1,41 @@
...
+      );
+
+      if (puzzleInstance.error) {
+        throw new Error(puzzleInstance.message);
+      } else {
+        setPuzzleInstanceData(puzzleInstance);
+      }
+    };
+    toast.promise(puzzleInstancePromise(), {
+      loading: 'Making your puzzle instance... ⚙️',
```
Opeyem1a, line 35: Should this auto-dismiss after a period of time?

src/pages/puzzles/[instanceId]/index.tsx
```diff
@@ -38,35 +41,55 @@ const PuzzlePage = ({
...
+          user
+        );
 
-      if ((submission as HandledError).error) {
-        // TODO: handle errors from submitting
-        alert(submission.message);
-        return;
-      }
+        if ((submission as HandledError).error) {
+          console.error((submission as HandledError).message);
```
Comment (Opeyem1a) [inline src/pages/puzzles/[instanceId]/index.tsx:54]: console?

src/pages/puzzles/[instanceId]/index.tsx
Comment (Opeyem1a) [reply src/pages/puzzles/[instanceId]/index.tsx:74]: But does this work? Did you test that it does what you think it does in every case?

src/pages/puzzles/[instanceId]/index.tsx
```diff
@@ -38,35 +41,55 @@ const PuzzlePage = ({
...
+        loading: 'Submitting your answer... ⏳',
+        success: 'Correct! Nice work!',
+        error: err => {
+          return err.message;
+        }
+      });
     } else {
-      alert('You must be logged in to submit puzzles!');
+      toast.error(t => (
+        <span>
```
Comment (Opeyem1a) [inline src/pages/puzzles/[instanceId]/index.tsx:81]: Yeah idk how I feel about having a bunch of custom HTML, can't we make components to put inside this? And just have a few different types of toast components

Opeyem1a (changes requested):

src/components/App/PuzzleGenerate/index.tsx
ketphan02: `toast.promise` will automatically change to success or fail

src/pages/puzzles/[instanceId]/index.tsx
Comment (ketphan02) [reply src/pages/puzzles/[instanceId]/index.tsx:54]: i think instead of sending the error through toaster for the end-user, I think it's better to send to console (`console.error` will send a message with error format so it's easy to identify)

src/pages/puzzles/[instanceId]/index.tsx
Comment (ketphan02) [reply src/pages/puzzles/[instanceId]/index.tsx:81]: i'll do that!

ketphan02:
Use on top right

src/pages/puzzles/[instanceId]/index.tsx
Comment (Opeyem1a) [reply src/pages/puzzles/[instanceId]/index.tsx:54]: But how would the user see the `console.error()`, wouldn't they have to already have developer tools open to see it?

Opeyem1a (approved):
This seems fine now - just grab me some images of what it looks like @ketphan02

ketphan02:
@Opeyem1a

https://user-images.githubusercontent.com/35695324/152636429-4740be90-6fb7-4afa-adff-cfb6b5aadf45.mov


https://user-images.githubusercontent.com/35695324/152636488-9eba9dcd-75ae-4d7e-8a44-f78b690101bf.mov

Opeyem1a:
Looks good!

PR #99, opened 2022-01-18
Add hash.js

ketphan02:


package.json
```diff
@@ -18,9 +18,12 @@
     "@types/react": "^17.0.27",
     "axios": "^0.24.0",
     "babel-plugin-superjson-next": "^0.4.2",
+    "hash.js": "^1.0.1",
+    "i": "^0.3.7",
     "next": "^12.0.7",
     "next-auth": "^4.0.0-beta.4",
     "next-transpile-modules": "^9.0.0",
+    "npm": "^8.3.1",
```
yuqi88, line 26: just a question I have here.
Are you trying to update npm? Line 22 & 26 mean that when we run npm install it will install "install" and "npm," and logically that doesn't make sense...?

Opeyem1a:
I have the same questions/comments as Lydia

package.json
ketphan02: Oh that's good point, let me fix this

Opeyem1a:
Is this PR dead because of @yuqi88's #101?
Close it if it's dead @ketphan02

PR #101, opened 2022-01-25
Radio button component2

### LINKED ISSUE #91: Unit Testing on functions in /utils

yuqi88:
- improve the style of the radio button when they're unchecked to all-gray.
- changed the hash library
- add unit tests to the util functions

close #91 , #92 , #100


src/__tests__/utils/password.test.ts
```diff
@@ -0,0 +1,18 @@
+import { checkHash, hashFunction } from '../../utils/password';
+
+describe('testing the hashFunction', () => {
+  it('should compare the password hashed and without', () => {
+    expect(hashFunction('password')).not.toBe('password');
+  });
+
+  it('should compare the length of 2 hashed passwords', async () => {
+    const hashed1 = await hashFunction('password1');
```
Opeyem1a, line 9: Might be even better to generate 2 strings with randomized length instead of hardcoding them

src/__tests__/utils/password.test.ts
```diff
@@ -0,0 +1,18 @@
+import { checkHash, hashFunction } from '../../utils/password';
+
+describe('testing the hashFunction', () => {
+  it('should compare the password hashed and without', () => {
```
Opeyem1a, line 4: this is a nitpick, but hashing is technically not exclusive to passwords, so we should use `'secret'` instead of `'password'`

src/__tests__/utils/password.test.ts
```diff
@@ -0,0 +1,18 @@
...
+  });
+
+  it('should compare the length of 2 hashed passwords', async () => {
+    const hashed1 = await hashFunction('password1');
+    const hashed2 = await hashFunction('pw2');
+    expect(hashed1.length).toBe(hashed2.length);
+  });
+});
+describe('testing the hashFunction', () => {
+  it('should return the same results', async () => {
```
Opeyem1a, line 15: this feels like a bad explanation of what this does, it's more so testing that the same secret always creates the same hashed value

src/__tests__/utils/password.test.ts
```diff
@@ -0,0 +1,18 @@
...
+    expect(hashFunction('password')).not.toBe('password');
+  });
+
+  it('should compare the length of 2 hashed passwords', async () => {
+    const hashed1 = await hashFunction('password1');
+    const hashed2 = await hashFunction('pw2');
+    expect(hashed1.length).toBe(hashed2.length);
+  });
+});
+describe('testing the hashFunction', () => {
```
Opeyem1a, line 14: Why did we make a separate `describe()` here?

src/__tests__/utils/puzzles.test.ts
```diff
@@ -0,0 +1,29 @@
+import { checkPuzzleAnswer } from '../../utils/puzzles';
+import { Puzzle } from '@prisma/client';
+
+it('should compare the user submitted puzzle answer with the ans key in the puzzle (from the db)', () => {
+  const puzzle = {
```
Opeyem1a, line 5: As discussed in the meeting, move object definitions like this into some kind of `/mocks` directory

src/__tests__/utils/puzzles.test.ts
```diff
@@ -0,0 +1,29 @@
...
+    variables: {
+      answer: '2'
+    },
+    puzzleType: {},
+    puzzleTypeId: 1,
+    puzzleInstances: []
+  } as Puzzle;
+
+  expect(checkPuzzleAnswer(puzzle, 1, '3')).toBeFalsy();
+  expect(checkPuzzleAnswer(puzzle, 1, '2')).toBeTruthy();
```
Opeyem1a, line 28: This is alright, but here's a new concept. Since we are creating fake objects in order to run these tests, it's good practice to not make the tests depend too heavily on the fake data. In other words, instead of hardcoding the `'2'` on line 28, you could put `expect(checkPuzzleAnswer(puzzle, 1, puzzle.variables.answer)).toBeTruthy();`, which logically performs the same test, but now we can change the fake object's answer anytime and the test doesn't need to be updated.

Lmk if you want to talk about this further, but hopefully ti makes sense

src/__tests__/utils/feedbackGif.test.ts
```diff
@@ -0,0 +1,22 @@
+import { getRandomGifSrc } from '../../utils/feedbackGif';
+
+const arr = [0, 1, 3];
```
Opeyem1a, line 3: the array passed into this function is meant to be a string[] - it's my bad for not adding proper types to the function's header (add them now pls)

It's supposed to be an array of filepaths like `['1.gif', '2.gif', '3.gif']`

src/__tests__/utils/feedbackGif.test.ts
```diff
@@ -0,0 +1,22 @@
...
+let isCorrect = true;
+describe('randomly generating the correct/incorrect gifs:', () => {
+  it('from the correct-gif & its path', () => {
```
Opeyem1a, line 6: I think it's best to check what the return value `startsWith()` in order to see if the `/feedbackGifs/correct/` or `/feedbackGifs/incorrect/` is the right one.

and then leave a single test with the `.toContain()` structure you have currently

Opeyem1a (changes requested):

src/__tests__/utils/puzzles.test.ts
Opeyem1a: Also smart way to generate the false value that references the real answer too

Opeyem1a: Switch `toBeTruthy` and `toBeFalsey` into `toBe(false)` and `toBe(true)`

src/utils/password.ts
```diff
@@ -1,11 +1,11 @@
...
-const hashFunction = (secret: string) => {
-  return sha256().update(secret).digest('hex');
+const hashFunction = async (secret: string) => {
+  return argon2.hash(secret, { type: argon2.argon2id });
 };
 
-const checkHash = (checker: string, hash: string) => {
-  return hashFunction(checker) === hash;
+const checkHash = async (checker: string, hash: string) => {
+  return (await hashFunction(checker)) === hash;
```
ketphan02, line 8: I think this will be
```ts
await argon2.verify(<hash>, <password>)
```

src/utils/password.ts
```diff
@@ -1,11 +1,11 @@
-import { sha256 } from 'hash.js';
+import argon2 from 'argon2';
 
-const hashFunction = (secret: string) => {
-  return sha256().update(secret).digest('hex');
+const hashFunction = async (secret: string) => {
+  return argon2.hash(secret, { type: argon2.argon2id });
```
ketphan02, line 4: I think this also need error handler since it's a promise (the same with `checkHash()`)

src/__tests__/utils/password.test.ts
```diff
@@ -0,0 +1,18 @@
...
+  it('should return the same results', async () => {
+    expect(checkHash('password', await hashFunction('password'))).toBeTruthy();
```
ketphan02, line 16: `toBe(true)`

ketphan02 (changes requested):

src/__tests__/utils/password.test.ts
yuqi88: Cool, I didn't know that.

src/__tests__/utils/password.test.ts
yuqi88: Good catch, the second describe is actually for the checkHash function. I'll change that.

src/__tests__/utils/password.test.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  it('should compare the length of 2 hashed passwords', async () => {
+    //add 2 random length passwords: 1-10
+    const len1 = Math.floor(Math.random()*10+1);
+    const len2 = Math.floor(Math.random()*10+1);
+    // generate random strings with the length above;
+    const pw1 = getRandomString(len1);
+    const pw2 = getRandomString(len2);
+    console.log("password 1 is " + pw1);
+    console.log("password 2 is " + pw2);
```
ketphan02, line 17: Remove this!

src/__tests__/utils/password.test.ts
```diff
@@ -0,0 +1,37 @@
...
+});
+
+const getRandomString = (length) =>{
+  const randomChars = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+  let str = '';
+  for (let i=0; i<length; i++){
+    str += randomChars.charAt(Math.floor(Math.random() * randomChars.length));
+  }
+  return str;
+}
```
ketphan02, line 37: use Faker for this

src/__tests__/utils/puzzles.test.ts
```diff
@@ -0,0 +1,18 @@
+import { checkPuzzleAnswer } from '../../utils/puzzles';
+import puzzle from '../mocks/mock_puzzle';
+
+it('should compare the user submitted puzzle answer with the ans key in the puzzle (from the db)', () => {
+  const answer = puzzle.variables.answer;
```
ketphan02, line 5: what if answer does not appear in variables ?

ketphan02 (changes requested):

src/__tests__/mocks/mock_puzzle.ts
```diff
@@ -0,0 +1,25 @@
+import {Puzzle} from "@prisma/client";
```
ketphan02, line 1: this file should not be inside __test__ folder, it should be in __mock__

package-lock.json
```diff
@@ -1,11126 +1,29 @@
 {
   "name": "project",
   "version": "0.1.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
   "requires": true,
```
ketphan02, line 5: Do not delete package-lock.json

ketphan02 (changes requested):

src/__tests__/utils/error.test.ts
```diff
@@ -0,0 +1,44 @@
+import { handleServiceError } from '../../utils/error';
+import { HandledError } from '../../types/error';
+
+describe('HTTP response status', () => {
+  it('is 400', () => {
```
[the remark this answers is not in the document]
Opeyem1a: Like the string? I think it's probably okay as-is, but if you have better suggestions feel free to drop them here

src/__tests__/utils/puzzles.test.ts
yuqi88: What do you mean? Isn't this the way how Puzzle is set up in the Prisma scheme? Are you suggesting that we shouldn't hard code that (line5) or you meant something else?

ketphan02: Because in schema, `variables` is an object, therefore, it cannot guarantee that there's a field named `answer`. You should check if `answer` exist in variables or use `puzzle.variables?.answer`

yuqi88: I see!

Opeyem1a:
I can't remember when I last reviewed this but I think our consensus was to resolve tests on this

src/__tests__/utils/puzzles.test.ts
```diff
@@ -0,0 +1,22 @@
...
+import { mockRandomSeed } from '../../__mocks__/pages/api/puzzles/instances/submit.ts';
+import {getRandomNumber, getRandomString} from '../../__mocks__/pages/api/getRandom';
+import puzzle from "../../__mocks__/mock_puzzle";
+
+it('should compare the user submitted puzzle answer with the ans key in the puzzle (from the db)', () => {
+  let pzl;
+  let ans;
+  try{
+    pzl = pzl = mockPuzzle();
+    ans = mockPuzzle().variables?.answer;
```
ketphan02, line 12: check if the mockPuzzle has variables, else, update the mockPuzzle to have variables

src/__tests__/utils/puzzles.test.ts
```diff
@@ -0,0 +1,22 @@
...
+  }catch (e){
+    console.log(e);
```
ketphan02, line 14: use console.error instead

src/__tests__/utils/puzzles.test.ts
```diff
@@ -0,0 +1,22 @@
...
+import {getRandomNumber, getRandomString} from '../../__mocks__/pages/api/getRandom';
+import puzzle from "../../__mocks__/mock_puzzle";
+
+it('should compare the user submitted puzzle answer with the ans key in the puzzle (from the db)', () => {
+  let pzl;
+  let ans;
+  try{
+    pzl = pzl = mockPuzzle();
+    ans = mockPuzzle().variables?.answer;
+  }catch (e){
```
ketphan02, line 13: needs spacing

src/__tests__/utils/puzzles.test.ts
```diff
@@ -0,0 +1,22 @@
...
+    console.log(e);
+  }finally {
```
ketphan02, line 15: this also needs spacing

src/__mocks__/mock_puzzle.ts
```diff
@@ -0,0 +1,25 @@
...
+  question: 'What is 1+1?',
+  inputType: 'TEXT',
+  published: false,
+  variables: {
+    answer: '2'
+  },
+  puzzleType: {},
+  puzzleTypeId: 1,
+  puzzleInstances: []
+} as Puzzle;
```
ketphan02, line 23: This still contains a lot of constants. Since my puzzles have been up, you can try to integrate that to your code

src/__mocks__/pages/api/getRandom/index.ts
```diff
@@ -0,0 +1,9 @@
+import * as faker from 'faker';
```
ketphan02, line 1: The file name should match with the functions you're working with

ketphan02: If not, bring it to the index of /api

src/__mocks__/pages/api/puzzles/index.ts
```diff
@@ -9,7 +9,7 @@ import {
   mockHint,
   mockLatitude,
   mockLongtitude,
-  mockName,
+  mockName, mockOfficialAnswer,
```
ketphan02, line 12: endline

src/__mocks__/pages/api/puzzles/index.ts
```diff
@@ -54,7 +54,9 @@ export const mockPuzzle = (): Promise<Puzzle> => {
       difficulty: mockDifficulty(),
       content: [mockContent(), mockContent(), mockContent()],
       question: mockQuestion(),
-      variables: {}, // Not sure what to put here
+      variables: {
+        answer: mockOfficialAnswer(),
```
ketphan02, line 58: Have not decided the type of puzzle (MCQ or TEXT), so should not put here

ketphan02: This can be done when calling the function later, this is like a skeleton

src/__mocks__/pages/api/puzzles/instances/create.ts
```diff
@@ -18,6 +18,9 @@ export const mockDifficulty = (): DIFFICULTY => {
   ] as DIFFICULTY;
 };
 
+export  const mockOfficialAnswer = (): string =>{
```
ketphan02, line 21: double space between export and const

src/__tests__/pages/api/auth/reset-password.test.ts
```diff
@@ -1,3 +1,4 @@
+import { PrismaClient } from '@prisma/client';
```
ketphan02, line 1: don't need this

src/__tests__/utils/password.test.ts
```diff
@@ -0,0 +1,31 @@
...
+    expect(await hashFunction('secret_hi')).not.toBe('secret_hi');
+    expect(await hashFunction('secret_hi')).not.toBe(
+      await hashFunction('secret_hi')
+    );
+  });
+
+  it('should compare the length of 2 hashed passwords', async () => {
+    //add 2 random length passwords: 1-10
+    const len1 = Math.floor(Math.random() * 10 + 1);
+    const len2 = Math.floor(Math.random() * 10 + 1);
```
ketphan02, line 15: Make another mock called randomNumber and assign it instead of hardcoding number 10

ketphan02 (changes requested):

ketphan02:
And do the versioning thing since there's a lot of change in `package-lock.json`

src/__tests__/utils/puzzles.test.ts
Opeyem1a: Nah I think we should just have a mock for MCQ in addition to a mock for text answer

yuqi88: 1. mockPuzzle is updated, so it does have "variables"
2. I kept the static mock puzzle in case the mockPuzzle fail to have values in variable or answer

src/__mocks__/mock_puzzle.ts
yuqi88: I didn't delete this as an alternative for mockPuzzle in case "variables" or the answer is empty.

yuqi88:
I have updated my node.js and npm to the latest version :)

src/utils/password.ts
```diff
@@ -1,11 +1,19 @@
...
 };
 
-const checkHash = (checker: string, hash: string) => {
-  return hashFunction(checker) === hash;
+const checkHash = async (secret: string, hashValue: string) => {
+  try {
+    return await argon2.verify(hashValue, secret);
+  } catch (err) {
+    throw err;
+  }
```
ketphan02, line 16: `argon2.verify` already a Promise, don't need to do async await here

src/utils/password.ts
```diff
@@ -1,11 +1,19 @@
...
+import argon2 from 'argon2';
 
-const hashFunction = (secret: string) => {
-  return sha256().update(secret).digest('hex');
+const hashFunction = async (secret: string) => {
+  try {
+    return argon2.hash(secret, { type: argon2.argon2id });
+  } catch (err) {
+    throw err;
+  }
```
ketphan02, line 8: `argon2.hash` already a Promise, don't need to do async await here

ketphan02 (changes requested):

src/__tests__/utils/puzzles.test.ts
```diff
@@ -0,0 +1,22 @@
...
+import { mockPuzzle } from '../../__mocks__/pages/api/puzzles';
+import { mockRandomSeed } from '../../__mocks__/pages/api/puzzles/instances/submit.ts';
+import {getRandomNumber, getRandomString} from '../../__mocks__/pages/api/getRandom';
+import puzzle from "../../__mocks__/mock_puzzle";
+
+it('should compare the user submitted puzzle answer with the ans key in the puzzle (from the db)', () => {
+  let pzl;
+  let ans;
+  try{
+    pzl = pzl = mockPuzzle();
```
ketphan02, line 11: seems not right

ketphan02: Also, mockPuzzle is a Promise,

src/__tests__/utils/puzzles.test.ts
```diff
@@ -0,0 +1,22 @@
+import { checkPuzzleAnswer } from '../../utils/puzzles';
+import { mockPuzzle } from '../../__mocks__/pages/api/puzzles';
+import { mockRandomSeed } from '../../__mocks__/pages/api/puzzles/instances/submit.ts';
+import {getRandomNumber, getRandomString} from '../../__mocks__/pages/api/getRandom';
+import puzzle from "../../__mocks__/mock_puzzle";
+
```
ketphan02, line 6: for every test suit, add a description

ketphan02 (changes requested):

src/__tests__/utils/puzzles.test.ts
```diff
@@ -0,0 +1,22 @@
+import { checkPuzzleAnswer } from '../../utils/puzzles';
+import { mockPuzzle } from '../../__mocks__/pages/api/puzzles';
+import { mockRandomSeed } from '../../__mocks__/pages/api/puzzles/instances/submit.ts';
```
ketphan02, line 3: Remove .ts

src/__tests__/utils/feedbackGif.test.ts
```diff
@@ -0,0 +1,21 @@
+import { getRandomGifSrc } from '../../utils/feedbackGif';
+
+//returns a path from a provided array
```
ketphan02, line 3: missing description

src/__tests__/utils/difficulty.test.ts
```diff
@@ -0,0 +1,7 @@
+import { difficultySentenceCase } from '../../utils/difficulty';
+
```
ketphan02, line 2: missing description

src/__tests__/pages/api/auth/reset-password.test.ts
```diff
@@ -1,3 +1,4 @@
...
 import { NextApiRequest, NextApiResponse } from 'next';
 import { PrismaAdapter } from '@next-auth/prisma-adapter';
```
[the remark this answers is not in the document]
ketphan02: Also, that's the wrong way to check the hash function, argon2 is more secured, so 1 string cannot be hashed to 1 hash (but more hashes)
must check if the same hash is different

ketphan02 (changes requested):
This will help you resolve all tests but 1 test about the answer, I think you have to figure this out

src/__tests__/utils/puzzles.test.ts
ketphan02: you have to await it

src/__tests__/utils/puzzles.test.ts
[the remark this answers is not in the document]
yuqi88: I'll change it for consistency, but I thought it works both ways? Or is`.variables['answer']`  better than `.variables.answer` ?

src/__tests__/utils/password.test.ts
```diff
@@ -0,0 +1,31 @@
+import { checkHash, hashFunction } from '../../utils/password';
+import {getRandomNumber, getRandomString} from '../../__mocks__/pages/api/getRandom';
+
+describe('testing the hashFunction', () => {
+  it('should compare the secret hashed and without', async () => {
+    expect(await hashFunction('secret_hi')).not.toBe('secret_hi');
+    expect(await hashFunction('secret_hi')).not.toBe(
+      await hashFunction('secret_hi')
```
ketphan02, line 8: use random string

ketphan02 (changes requested):

src/__mocks__/mock_puzzle.ts
yuqi88: okay I deleted this

src/__mocks__/getRandom/index.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export const getRandomNumberRange = (min:number, max:number ):number => {
+    return faker.datatype.number({min: min, max: max});
+};
+
+export const getRandomNumber = ():number => {
+    return faker.datatype.number();
+};
+
+export const getRandomNumArray = (length?:number) => {
```
Opeyem1a, line 22: A better way to do this is to set a default value for the length parameter when it's not given:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters

That way it always runs the same code (it's a bit overkill to have it also randomize the length of the array when specified)

src/__mocks__/pages/api/puzzles/instances/create.ts
```diff
@@ -18,6 +19,15 @@ export const mockDifficulty = (): DIFFICULTY => {
   ] as DIFFICULTY;
 };
 
+export const mockOfficialAnswer = (): string => {
+  return faker.datatype.string();
+};
+
+export const mockOptionsNoAns = (): string[] => {
+  const str = getRandomString(getRandomNumberRange(1,4));
+  return  Array.from(str);
```
Opeyem1a, line 28: prettify

src/__tests__/utils/difficulty.test.ts
```diff
@@ -0,0 +1,14 @@
...
+describe('Difficulties are in lowercase except for the first letter.', ()=>{
+  it('Difficulty = EASY correct ', () => {
```
Opeyem1a, line 4: This goes for all tests:
In general, you want these tests to read like english sentences, this is for convention and also for others understanding it later. So `it('properly sentence cases MEDIUM difficulty', ...)` is better here in my opinion

src/__tests__/utils/error.test.ts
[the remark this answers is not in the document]
Opeyem1a: Yeah I was tripping before, this is better like `it('handles 400 errors', ...)`
(and that applies to the whole file)

src/__tests__/utils/password.test.ts
```diff
@@ -0,0 +1,25 @@
...
+import { getRandomString } from '../../__mocks__/getRandom';
+
+describe('testing the hashFunction: ', () => {
+  it('should compare the secret hashed and without', async () => {
+    const secret = getRandomString();
+    expect(await hashFunction(secret)).not.toBe(secret);
+    expect(await hashFunction(secret)).not.toBe(await hashFunction(secret));
+  });
+
+  it('should compare the length of 2 hashed passwords', async () => {;
```
Opeyem1a, line 11: these test strings describe what the test is doing but they should describe the desired outcome of the test.
i.e. instead of `'should compare the secret hashed and without'` it should be `'should not return the same hash function twice for the same string'`

i.e. `'should compare the length of 2 hashed passwords'` should be `'should return hashed strings of the same length regardless of the strings passed'`

src/__tests__/utils/password.test.ts
```diff
@@ -0,0 +1,25 @@
+import { checkHash, hashFunction } from '../../utils/password';
+import { getRandomString } from '../../__mocks__/getRandom';
+
+describe('testing the hashFunction: ', () => {
+  it('should compare the secret hashed and without', async () => {
+    const secret = getRandomString();
+    expect(await hashFunction(secret)).not.toBe(secret);
```
Opeyem1a, line 7: this seems like it's testing just that the hash function works at all, which sounds like it should be a separate test

src/__tests__/utils/password.test.ts
```diff
@@ -0,0 +1,25 @@
...
+  it('should compare the length of 2 hashed passwords', async () => {;
+    const pw1 = getRandomString();
+    const pw2 = getRandomString();
+    const hashed1 = await hashFunction(pw1);
+    const hashed2 = await hashFunction(pw2);
+    expect(hashed1.length).toBe(hashed2.length);
+  });
+});
+
+describe('testing the checkHash: ', () => {
```
Opeyem1a, line 20: `describe('validating a hash string', ...)`

src/__tests__/utils/puzzles.test.ts
```diff
@@ -0,0 +1,57 @@
+import { checkPuzzleAnswer } from '../../utils/puzzles';
+import { mockPuzzle } from '../../__mocks__/pages/api/puzzles';
+import { mockRandomSeed } from '../../__mocks__/pages/api/puzzles/instances/submit';
+import {getRandomString} from '../../__mocks__/getRandom';
+import {prisma} from "../../__mocks__";
+import {mockOfficialAnswer, mockOptionsNoAns} from "../../__mocks__/pages/api/puzzles/instances/create";
+
+describe('testing checkPuzzleAnswer: Comparing user submitted answer with the ans key in the puzzle (from the db) - ', () => {
```
Opeyem1a, line 8: from past comments, fix the strings in `describe` and `it`

src/utils/puzzles.ts
```diff
@@ -6,6 +6,7 @@ const checkPuzzleAnswer = (
   puzzleAnswer: string
 ) => {
   const trueAnswer = puzzle.variables['answer'];
+  console.log("in the checkPuzzleAnswer function, the accessed ans = " + trueAnswer);
```
Opeyem1a, line 9: remove

Opeyem1a (changes requested):

Opeyem1a (changes requested):

Opeyem1a (approved):

PR #102, opened 2022-01-28
Add Jest for API testing and test for Sign Up page

ketphan02:
+ Add a success test for sign up function.
+ Discuss about library to use to fetch api from backend
+ Close #104

PR #105, opened 2022-01-30
Reset Password test add

### LINKED ISSUE #106: Test reset-password api endpoint

ketphan02:
* Add test for password reset feature

Need to merge #102 before merging this
Close #106


jest.setup.ts
```diff
@@ -1 +1,3 @@
 import '@testing-library/jest-dom/extend-expect'
+
+jest.setTimeout(30000);
```
Opeyem1a, line 3: Why?

package.json
```diff
@@ -8,7 +8,7 @@
     "build": "next build",
     "start": "npx prisma generate && next start",
     "lint": "next lint",
-    "test": "jest --watch",
+    "test": "docker-compose up -d db && npx prisma migrate dev && jest && docker-compose down",
```
Opeyem1a, line 11: Does it destroy the test db though?

src/__mocks__/pages/api/auth/index.ts
```diff
@@ -0,0 +1,32 @@
+import * as faker from 'faker';
+
+export type UserDataProp = {
+  email: string;
+  password: string;
+};
+
+export const mockEmail = () => {
+  return faker.internet.password();
```
Opeyem1a, line 9: isn't this supposed to be an email?

src/pages/api/auth/signup.ts
```diff
@@ -20,7 +20,7 @@ const signUpHandler = async (req: NextApiRequest, res: NextApiResponse) => {
   try {
     const hashPassword = hashFunction(password);
     const user = await signUp({ email, password: hashPassword });
-    res.status(200).json(user);
+    res.status(201).json(user);
```
Opeyem1a, line 23: Good catch

docker-compose.yml
```diff
@@ -32,4 +32,4 @@ volumes:
   db_data:
 
 networks:
-  app_network:
\ No newline at end of file
+  app_network:
```
Opeyem1a, line 35: Is this just space?

src/__mocks__/pages/api/auth/index.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const mockUserData = () => {
+  return {
+    email: mockEmail(),
+    password: mockPassword()
+  } as UserDataProp;
+};
+
+export const mockManyUserData = (numUsers: number) => {
+  const arr: UserDataProp[] = [];
```
Opeyem1a, line 24: nitpick but let's use better variable names than `arr`

src/__tests__/pages/api/auth/reset-password.test.ts
```diff
@@ -0,0 +1,299 @@
...
+  const req = {
+    body: {
+      email,
+      password
+    }
+  } as NextApiRequest;
+
+  const json = jest.fn();
+
+  const status = jest.fn(() => {
```
Opeyem1a, line 34: I don't understand how this works so you'll have to explain it to me so we can confirm the pattern

src/__tests__/pages/api/auth/reset-password.test.ts
```diff
@@ -0,0 +1,299 @@
...
+    return { json };
+  });
+
+  const res = {
+    status
+  } as unknown as NextApiResponse;
```
Opeyem1a, line 40: as unknown?

src/__tests__/pages/api/auth/reset-password.test.ts
```diff
@@ -0,0 +1,299 @@
...
+
+  await signUpHandler(req, res);
+});
+
+describe('Successfully change password', () => {
+  it('Change password successs', async () => {
+    const hashPassword = (
+      await prisma.user.findUnique({
```
Opeyem1a, line 48: Logically this breaks if sign-up breaks though, right? When we sign up the mock user in the `beforeEach`, we should do it manually without the API endpoint that does it

src/__tests__/pages/api/auth/reset-password.test.ts
```diff
@@ -0,0 +1,299 @@
...
+        where: {
+          email
+        },
+        select: {
+          password: true
+        }
+      })
+    ).password;
```
Opeyem1a, line 56: Maybe a `?.`

src/__tests__/pages/api/auth/reset-password.test.ts
```diff
@@ -0,0 +1,299 @@
...
+    expect(hashFunction(newPassword)).toEqual(currentPassword);
+  });
+});
+
+describe('Fail to change password', () => {
+  it('Wrong oldPassword', async () => {
+    const req = {
+      body: {
+        email: email,
+        oldPassword: 'wrongPassword',
```
Opeyem1a, line 108: Hmmm but technically it's possible for this to actually be the right password though, right? Maybe do like `'wrongPassword' + password` to ensure it's always wrong

src/__tests__/pages/api/auth/reset-password.test.ts
```diff
@@ -0,0 +1,299 @@
...
+    const res = {
+      status
+    } as unknown as NextApiResponse;
+
+    await resetPasswordHandler(req, res);
+
+    expect(status.mock.calls[0]).toEqual([400]);
+    expect(json.mock.calls[0]).toEqual([
+      {
+        message: 'Old password is incorrect'
```
Opeyem1a, line 128: Make a card to make constants for our API response messages

src/__tests__/pages/api/auth/reset-password.test.ts
```diff
@@ -0,0 +1,299 @@
...
+      }
+    ]);
+  });
+
+  it('Email not inside db', async () => {
```
Opeyem1a, line 133: shouldn't it be like `it('Should not reset password of nonexistent user')...`

Opeyem1a (changes requested):
I didn't finish reviewing but a lot of the comments are repeats so I'll let you resolve the current ones first

src/__tests__/pages/api/auth/reset-password.test.ts
ketphan02: since the hash function is working (by unit testing the hash function), we can assume all password has the same length = 32 or 64 or sth i forgot haha

src/__mocks__/pages/api/auth/index.ts
ketphan02: oh shoot 🙈

jest.setup.ts
ketphan02: some test that forgot to resolve async or to close the db can run forever

package.json
Opeyem1a: Oh but I mean in terms of data, does the test db persist? It seems not worth to have past data from tests interfere with new tests. I think every time we run tests the test db should be empty

jest.setup.ts
Opeyem1a: Oh then this is cool

src/__tests__/pages/api/auth/reset-password.test.ts
Opeyem1a: But when we send `oldPassword` to the API endpoint, it'll hash it anyway and compare by hashed value. Technically, the original unhashed password could be `'wrongPassword'` since it's random, right? If yes, then we need a way to make sure

docker-compose.yml
ketphan02: I added an end of line, it was an end of file there

src/__tests__/pages/api/auth/reset-password.test.ts
ketphan02: I got error by typescript when I tried without unknown,
```txt
Type '{ status: Mock<{ json: Mock<any, any>; }, []>; }' is missing the following properties from type 'ServerResponse': statusCode, statusMessage, assignSocket, detachSocket, and 54 more.
```
this indicates that a lot of properties of type `ServerResponse` are not defined, and it's efficient to not define them. Adding `unknown` is like telling it's an unstrict type `ServerResponse`

src/__tests__/pages/api/auth/reset-password.test.ts
ketphan02: oh yes!

Opeyem1a (approved):

PR #108, opened 2022-01-30
Setup DB

### LINKED ISSUE #107: Find a way to setup db and use test environment

ketphan02:
Close #107

### DIFF
--- a/jest.setup.ts
+++ b/jest.setup.ts
@@ -1,1 +1,5 @@
 import '@testing-library/jest-dom/extend-expect'
+import dotenv from 'dotenv';
+dotenv.config({
+  path: '.env.test',
+});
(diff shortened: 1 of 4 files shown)


jest.setup.ts
```diff
@@ -1 +1,5 @@
 import '@testing-library/jest-dom/extend-expect'
+import dotenv from 'dotenv';
+dotenv.config({
+  path: '.env.test',
```
Opeyem1a, line 4: So this works now? Like it automatically uses `.env.test` when testing?

Opeyem1a (changes requested):
Just double checking that this works, otherwise it's good to go

jest.setup.ts
ketphan02: That's right!

Opeyem1a (approved):

PR #110, opened 2022-02-09
Admin and User test

ketphan02:
- [x] Admin validation test
- [x] User retrieve test


src/__tests__/pages/api/admin/validate.test.ts
```diff
@@ -0,0 +1,116 @@
...
+      id: expect.any(Number),
+      image: null,
+      name: null,
+      password: expect.any(String),
+      updatedAt: expect.any(Date)
+    });
+  });
+});
+
+describe('/api/admin/validate: Missing parameters for API', () => {});
```
Opeyem1a, line 116: What does this do?

src/__tests__/pages/api/admin/validate.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    const res = {
+      status
+    } as unknown as NextApiResponse;
+
+    await adminValidate(req, res);
+
+    expect(json).toHaveBeenCalledTimes(1);
+    expect(status).toHaveBeenCalledTimes(1);
+    expect(status).toHaveBeenCalledWith(403);
+    expect(json).toHaveBeenCalledWith({
```
Opeyem1a, line 103: Conceptually, what is this testing?

src/__tests__/pages/api/admin/validate.test.ts
```diff
@@ -0,0 +1,116 @@
...
+beforeEach(async () => {
+  await prisma.user.deleteMany();
+  await prisma.account.deleteMany();
+});
+
+describe('/api/admin/validate: Success', () => {
+  const adminEmail = mockEmail();
+
+  it('Sucessfully validate admin', async () => {
+    const adminUser = await createUser({
```
Opeyem1a, line 22: I think this is probably fine for now, but it'd be nice of all these standard tasks like "create admin user with this email" were just functions. Almost like a "test utils" folder. I think it's good in the long run to have the code in tests explicitly be the test itself and not lots of extra set up code that makes files longer than they need to be. Anyway, if this makes sense make a card to do this on Trello and resolve this conversation

Opeyem1a (changes requested):

src/__tests__/pages/api/admin/validate.test.ts
ketphan02: oh it not yet done haha, they're fail tests

ketphan02:
Need to wait until #102 to be merged

src/__tests__/pages/api/admin/validate.test.ts
ketphan02: I still not get this, please explain in the meeting @Opeyem1a

src/__tests__/pages/api/admin/validate.test.ts
ketphan02: My bad my bad, it shouldn't be like this, I tried a better way

src/__tests__/pages/api/auth/reset-password.test.ts
```diff
@@ -65,12 +63,10 @@ describe('Successfully change password', () => {
...
     await resetPasswordHandler(req, res);
 
-    expect(status.mock.calls[0]).toEqual([200]);
-    expect(json.mock.calls[0]).toEqual([
-      {
-        message: 'Password changed'
-      }
-    ]);
+    expect(status).toHaveBeenNthCalledWith(1, 200);
+    expect(json).toHaveBeenNthCalledWith(1, {
```
Opeyem1a, line 67: What do all these `1`s mean?

src/__tests__/pages/api/auth/signup.test.ts
```diff
@@ -1,32 +1,21 @@
...
   mockUserData,
   mockManyUserData
 } from '../../../../__mocks__/pages/api/auth';
 import type { UserDataProp } from '../../../../__mocks__/pages/api/auth';
-import { NextApiRequest, NextApiResponse } from 'next';
 
 const prisma = new PrismaClient();
 
-beforeAll(async () => {
-  await prisma.$connect();
```
Opeyem1a, line 19: We don't need the `$connect`/`$disconnect` anymore?

Opeyem1a (changes requested):

package.json
```diff
@@ -8,7 +8,7 @@
     "build": "next build",
     "start": "npx prisma generate && next start",
     "lint": "next lint",
-    "test": "jest",
+    "test": "jest --runInBand",
```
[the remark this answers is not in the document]
ketphan02: This make tests run asynchronously
For example, if you are testing creating user and testing creating N multiple users, when you retrive data from the database, it going to be N + 1 users and both tests might fail, `--runInBand` prevents this.

src/__tests__/pages/api/auth/reset-password.test.ts
ketphan02: Normaly, when sending the requests, you usualy expect it to send 1 time and give you the respond 1 time. The 1's here can spot out if there's any abnormalities.

src/__tests__/pages/api/auth/signup.test.ts
ketphan02: Since we have `--runInBand` and we reset the database after every run, we can get rid of it (I'm not sure about this lol, I removed but it still works)

Opeyem1a (changes requested):

ketphan02:
Will have to pull #116 before merging

src/__tests__/pages/api/auth/reset-password.test.ts
```diff
@@ -214,29 +191,14 @@ describe('/api/auth/reset-password: Failed', () => {
...
 
     await resetPasswordHandler(req, res);
 
-    expect(status.mock.calls[0]).toEqual([400]);
-    expect(json.mock.calls[0]).toEqual([
-      {
-        message: 'Email, password and new password are required'
-      }
-    ]);
+    expect(status).toHaveBeenNthCalledWith(2, 400);
```
Opeyem1a, line 198: What is the argument `2` supposed to be here

Opeyem1a (changes requested):

src/__tests__/pages/api/auth/reset-password.test.ts
ketphan02: I this test, I call the mock function 2 times, 1 with null para and 1 with undefined para. Therefore, we should expect this mock call function has been called 2 times and returns status 400 both times

Opeyem1a:
Just gotta resolve these conflicts @ketphan02

ketphan02:
@Opeyem1a all done!

PR #111, opened 2022-03-01
Puzzles tests

ketphan02:
+ Add tests for 6 APIs
+ Need to implement fail tests for some files

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -5,7 +5,7 @@ datasource db {
 
 generator client {
   provider = "prisma-client-js"
-  // previewFeatures = ["referentialIntegrity"]
+  previewFeatures = ["referentialIntegrity"]
 }
 
 model Account {
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -5,6 +5,7 @@ datasource db {
 
 generator client {
   provider = "prisma-client-js"
+  previewFeatures = ["referentialIntegrity"]
 }
 
 model Account {
@@ -96,7 +97,7 @@ model Puzzle {
   inputType       InputType        @default(TEXT) @map(name: "input_type")
   published       Boolean          @default(false)
   variables       Json
-  puzzleType      PuzzleType       @relation(fields: [puzzleTypeId], references: [id])
+  puzzleType      PuzzleType       @relation(fields: [puzzleTypeId], references: [id], onDelete: Cascade)
   puzzleTypeId    Int              @map(name: "puzzle_type_id")
   puzzleInstances PuzzleInstance[]
 
@@ -105,7 +106,7 @@ model Puzzle {
 
 model PuzzleInstance {
   id          Int          @id @default(autoincrement())
-  puzzle      Puzzle       @relation(fields: [puzzleId], references: [id])
+  puzzle      Puzzle       @relation(fields: [puzzleId], references: [id], onDelete: Cascade)
   puzzleId    Int          @map(name: "puzzle_id")
   hint        String       @default("No hint")
   submissions Submission[]
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -5,7 +5,7 @@ datasource db {
 
 generator client {
   provider = "prisma-client-js"
-  previewFeatures = ["referentialIntegrity"]
+  // previewFeatures = ["referentialIntegrity"]
 }
 
 model Account {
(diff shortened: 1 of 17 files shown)


ketphan02:
Will have to pull #116 before merging

prisma/schema.prisma
```diff
@@ -5,6 +5,7 @@ datasource db {
 
 generator client {
   provider = "prisma-client-js"
+  previewFeatures = ["referentialIntegrity"]
```
Opeyem1a, line 8: oh what's this?

src/__mocks__/pages/api/puzzles/instances/create.ts
```diff
@@ -0,0 +1,42 @@
+import * as faker from 'faker';
+import { DIFFICULTY } from '../../../../../types/global';
+
+const DIFFICULTY_LIST = ['EASY', 'MEDIUM', 'HARD'];
```
Opeyem1a, line 4: hmm, this creates 2 points of truth, but I know why you did it, leave a TODO so later we can figure out how to ensure this list is the same as the type definition

src/__mocks__/pages/api/puzzles/instances/submit.ts
```diff
@@ -0,0 +1,9 @@
+import * as faker from 'faker';
+
+export const mockAnswer = (): string => {
+  return faker.random.word();
```
Opeyem1a, line 4: this should optionally take a puzzle and if a puzzle is given that is MCQ type, it should randomly return one of the MCQ options. If the puzzle is not MCQ type, it should return the puzzles correct answer 50% of the time.

Problem with current implementation is it will essentially *never* be the right answer

Opeyem1a (changes requested):

src/__mocks__/pages/api/puzzles/index.ts
```diff
@@ -0,0 +1,70 @@
...
+      } : {
+        create: {
+          name: mockName(),
+          difficulty: mockDifficulty(),
+          content: [mockContent(), mockContent(), mockContent()],
+          question: mockQuestion(),
+          variables: {},
+          isGenerated: mockBoolean(),
+          puzzleType: {
+            create: {
```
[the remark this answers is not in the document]
ketphan02: if then, you also have to connect to this by
```js
puzzleType: {
    connect: mockPuzzleType.id
}
```

I think this is really similar to this create funciton

src/__mocks__/pages/api/puzzles/instances/create.ts
ketphan02: Sure!

prisma/schema.prisma
ketphan02: I think it comes with `onDelete: Cascade`

> In [3.0.1](https://github.com/prisma/prisma/releases/tag/3.0.1) and later it is possible to do cascading deletes using [Referential actions](https://www.prisma.io/docs/concepts/components/prisma-schema/relations/referential-actions). Be sure to follow the [upgrade guide](https://www.prisma.io/docs/guides/upgrade-guides/upgrading-versions/upgrading-to-prisma-3/referential-actions) before using this feature.

src/__mocks__/pages/api/puzzles/instances/submit.ts
ketphan02: You're right, I forgot there's another type of inputType!

ketphan02: The tests for this will be in submit.test.ts, not really what you said but I tested 2 kinds of inputTyoe

src/__mocks__/pages/api/puzzles/index.ts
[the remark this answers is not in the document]
Opeyem1a: Oh hmm

src/__mocks__/pages/api/puzzles/instances/submit.ts
Opeyem1a: Cool

prisma/schema.prisma
```diff
@@ -5,6 +5,7 @@ datasource db {
 
 generator client {
   provider = "prisma-client-js"
+  // previewFeatures = ["referentialIntegrity"]
```
Opeyem1a, line 8: Oh wait we commented it now?

Opeyem1a (approved):

prisma/schema.prisma
ketphan02: oh wait my bad

ketphan02: Resolved! Can you merge this please @Opeyem1a

PR #112, opened 2022-03-06
Add a script to dumps DB (puzzles only)

### LINKED ISSUE #94: Look into dumping/uploading DB data

ketphan02:
Closes #94


ketphan02:
Some changes were caused by running `npm run prettify`, I think we need to implement [husky](https://www.npmjs.com/package/husky), this helps format through `.prettierrc.js`

Opeyem1a:
I think it's something to make a card about and do later - doesn't seem critical to me, but I could be misunderstanding the problem @ketphan02

prisma/seed.ts
```diff
@@ -0,0 +1,95 @@
+import * as dotenv from 'dotenv';
+dotenv.config();
+
+import { PrismaClient, Prisma } from '@prisma/client';
+import * as faker from 'faker';
+
+const prisma = new PrismaClient();
+
+const main = async () => {
```
Opeyem1a, line 9: This is much closer to what I meant, but I still think the actual hardcoded data should not be in this file, it should still be an external `.json`, is that not possible?

Like it should be:
```ts
PUZZLE_TYPES.forEach(puzzleType => {
  const puzzleType = await prisma.puzzleType.create({
    data: {
      name: puzzleType.name
    }
  });
};
```

Where `PUZZLE_TYPES` is an array of objects defined in a separate file and imported here

Opeyem1a (changes requested):

README.md
```diff
@@ -24,6 +24,10 @@ Then run the following lines in the terminal
 npm i
 npm run dev
 ```
+If you want to have some puzzle without inputing every of them manually, run:
+```bash
+npx prisma db seed
```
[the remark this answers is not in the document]
ketphan02: it will trigger the dump function

package.json
```diff
@@ -34,7 +34,7 @@
   "devDependencies": {
     "@testing-library/jest-dom": "^5.14.1",
     "@testing-library/react": "^12.1.1",
-    "@types/faker": "^6.6.9",
+    "@types/faker": "^5.5.9",
```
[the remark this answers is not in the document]
ketphan02: faker.js 's creator has stopped maintaining the package, other packages are not as good as the latest working version so i think we should stick with this until something better rises

package.json
```diff
@@ -53,5 +53,8 @@
     "supertest": "^6.2.2",
     "ts-jest": "^27.0.6",
     "typescript": "^4.4.4"
+  },
+  "prisma": {
+    "seed": "ts-node --compiler-options {\"module\":\"CommonJS\"} prisma/seed.ts"
```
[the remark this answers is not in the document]
ketphan02: it helps the script use import/export instead of require ones

prisma/seed.ts
Opeyem1a: Yep yep!

Opeyem1a:
Yikers tests are failing

prisma/seed/index.ts
```diff
@@ -0,0 +1,88 @@
...
+dotenv.config();
+
+import { PrismaClient, Puzzle, PuzzleType } from '@prisma/client';
+import * as faker from 'faker';
+
+const prisma = new PrismaClient();
+
+import seedData from './seed.json';
+import type { DIFFICULTY } from '../../src/types/global';
+console.log(seedData);
```
Opeyem1a, line 11: remove

prisma/seed/seed.json
```diff
@@ -0,0 +1,85 @@
+{
```
Opeyem1a, line 1: Niice, make me a trello card to update this

prisma/seed/index.ts
```diff
@@ -0,0 +1,88 @@
...
+    );
+    await prisma.puzzleInstance.create({
+      data: {
+        puzzle: {
+          connect: {
+            id: puzzle.id
+          }
+        },
+        hint: puzzleInstance.hint,
+        longitude: parseFloat(faker.address.longitude()),
```
Opeyem1a, line 74: Hmm does this need to be fake? Couldn't we define these too?

Opeyem1a (changes requested):

prisma/seed/index.ts
ketphan02: oh you want these to be defined too, i'll do that

ketphan02: I thought you want this haha

ketphan02:
The locations:
image

prisma/seed/index.ts
Opeyem1a: Actually yeah I guess I'll put real data here so it doesn't really matter

Opeyem1a (approved):

PR #117, opened 2022-03-20
Profile page styling

### LINKED ISSUE #96: Fix profile styling

ketphan02:
Closes #96 
- Add styling for profile page
image
image
- Add function for checking if user has password to change
- Add function and tests for updating username
- Need further work with birthyear


prisma/schema.prisma
```diff
@@ -47,7 +47,7 @@ model User {
   id            Int          @id @default(autoincrement())
   email         String?      @unique
   password      String?
-  name          String?
+  name          String?      @unique
```
Opeyem1a, line 50: why do we need this to be unique? We find users by email primarily

src/pages/api/user/update-username.ts
```diff
@@ -0,0 +1,46 @@
...
+  try {
+    const duplicateUser = await prisma.user.findUnique({
+      where: {
+        name: username
+      }
+    });
+
+    if (duplicateUser) {
+      return res.status(400).json({
+        message: 'Username already exists.'
```
Opeyem1a, line 30: Yeah idk if we need this though - or did you do this because the stats list usernames and we need to be able to tell who's who?

src/pages/auth/profile/index.tsx
```diff
@@ -2,93 +2,195 @@ import React, { useEffect } from 'react';
...
+      );
+    } catch (err) {
+      console.error(err);
+    }
   };
 
+  useEffect(() => {
+    if (!session) return;
+
+    console.log(session)
```
Opeyem1a, line 47: Remove

src/pages/auth/profile/index.tsx
```diff
@@ -2,93 +2,195 @@ import React, { useEffect } from 'react';
...
+      .then((user) => {
+        if (user?.password) setHasPassword(true);
+        if (user.name) setUsername(user.name)
+      })
+      .catch((err) => {
+        console.error(err);
+      });
+  }, [userEmail]);
+
+  const handleUpdateUsername = async () => {
```
Opeyem1a, line 70: I think there's too much logic in this page, we should move utility functions to `util` files

src/pages/auth/profile/index.tsx
```diff
@@ -2,93 +2,195 @@ import React, { useEffect } from 'react';
...
+              <div className={styles.profileDetailsInfo}>
+                <span>Email:</span>
+                <div>{userEmail}</div>
+              </div>
+              <div className={styles.profileDetailsInfo}>
+                <span>Username:</span>
+                <Input
+                  type="text"
+                  id="username"
+                  placeholder={username || ''}
```
Opeyem1a, line 124: Maybe like "Enter username..." by default instead of empty string - what do you think?

src/services/getUserByEmail.ts
```diff
@@ -0,0 +1,18 @@
+import { User } from '@prisma/client';
+import Axios from './axios';
+
+export const getUserByEmail = async (data: {
```
Opeyem1a, line 4: tbh I'm not sure what the difference is supposed to be between a service and a util, but we should agree on it - reply with what would you define them as and then we can discuss?

src/styles/pages/profile.module.scss
```diff
@@ -0,0 +1,60 @@
...
+    .profileDetails {
+      display: flex;
+      flex-direction: row;
+      gap: 2rem;
+      margin-bottom: 3rem;
+
+      .imgPlaceholder {
+        width: 200px;
+        height: 200px;
+        background-color: rgb(96, 96, 96);
```
Opeyem1a, line 23: ideally this should be a colour defined in variables

Opeyem1a (changes requested):
There's also a lot of other code in this PR, try merging main again and hopefully some of it goes away - I'm assuming I need to merge #110 before this?

prisma/schema.prisma
ketphan02: I thought we think of building a leader board or something ?

src/services/getUserByEmail.ts
ketphan02: I think the difference between util and service is util is used to fetch the API when util is just doing things that don't have to be rendered

src/pages/auth/profile/index.tsx
ketphan02: I don't think this is too much

ketphan02: And these functions are also responsible for rendering, therefore, if we want to make this simpler, we can break them down to smaller components

Opeyem1a: Hmm i guess now that I've taken a deeper look it's probably okay

prisma/schema.prisma
Opeyem1a: Ask about this in the Discord to see what everyone thinks

Opeyem1a (changes requested):

src/pages/auth/profile/index.tsx
```diff
@@ -1,94 +1,203 @@
...
   }, [status]);
 
+  useEffect(() => {
+    getUserByEmail({ email: userEmail })
+      .then(user => {
+        if (user?.password) setHasPassword(true);
+        if (user.name) setUsername(user.name);
+      })
+      .catch(err => {
+        console.error(err);
```
Opeyem1a, line 48: Shouldn't we toast the error properly?

src/pages/auth/profile/index.tsx
```diff
@@ -1,94 +1,203 @@
...
+          error: err => `Error: ${err.message}`
+        }
+      );
+    } catch (err) {
+      console.error(err);
+    }
+  };
+
+  const handleUpdateUsername = async () => {
+    // TODO: Regex for username
```
Opeyem1a, line 74: Make a trello card

src/pages/auth/profile/index.tsx
```diff
@@ -1,94 +1,203 @@
...
+                <span>Birth year:</span>
+                {/* TODO: Load birthyear */}
+                <Input
+                  type="number"
+                  id="birthyear"
+                  placeholder={''}
+                  required={false}
+                />
+                <Button
+                  onClick={() => toast('Place holder for update birthyear')}
```
Opeyem1a, line 152: Should say like "Coming soon!" since it's possible a user will see this. Leave todo comments and dev notes for actual TODO comments

Opeyem1a (changes requested):

src/pages/auth/profile/index.tsx
ketphan02: Sure!

src/__tests__/pages/api/user/update-username.test.ts
```diff
@@ -0,0 +1,156 @@
...
+
+    const status = jest.fn(() => {
+      return { json };
+    });
+
+    const res = {
+      status
+    } as unknown as NextApiResponse;
+
+    let req = {
```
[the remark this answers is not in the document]
ketphan02: It will be different in different cases, I don't think we should do this @Opeyem1a

src/pages/auth/profile/index.tsx
ketphan02: My bad

src/__tests__/pages/api/user/update-username.test.ts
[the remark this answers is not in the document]
Opeyem1a: I think the testing could be generalized, but I'll leave this point for now until I can look at the entire repo again

src/pages/auth/profile/index.tsx
```diff
@@ -1,94 +1,203 @@
...
+          newPassword: password
+        }),
+        {
+          loading: 'Changing password',
+          success: 'Password changed successfully',
+          error: err => `Error: ${err.message}`
+        }
+      );
+    } catch (err) {
+      console.error(err);
```
Opeyem1a, line 69: Actually wait do we ever need to console errors then? I left these alone on my last review because I thought you wanted to both toast and log errors

Opeyem1a (changes requested):

src/pages/auth/profile/index.tsx
ketphan02: It's might be better to log it into a file, I'll make a Trello card

ketphan02: I'll change that to toast

Opeyem1a (approved):

PR #121, opened 2022-03-29
Admin header

### LINKED ISSUE #120: Add header for admin page

yuqi88:
- Added AdminHeader component
- Can be used for user profile page too after modifying if we need that
- Added the component to the Admin page

close #120


Opeyem1a:
Could you attach a screenshot of this change?

Opeyem1a (changes requested):

ketphan02:
I thought we don't need this anymore?

PR #126, opened 2022-04-05
Improve Header Component

### LINKED ISSUE #125: Improve header component

yuqi88:
- Add a "sign-in" button in the header on the homepage when user isn't logged in
- Change the component name: components> global > Header to components> global > Navbar

close #125


yuqi88:
[Discuss] Login page VS Signup page 
be the default page when users press Get Started/Start Hunting/Learn More?

For now, the default is login, and I agree to leave it as it is now.
Reasoning: 
(1)Group 1: ppl who often come to the website will be the ones who are doing the puzzles (aka registered users). 
(2)Group 2: ppl who are new to the website (aka not having an account) will be a one-time thing, and they become the former group after the first visit.
If we set the default to sign-up, it's gonna be inconvenient in general for the main users of the website every time when they do the puzzles (assuming they have the habit of forgetting to login)

Other opinions?

Opeyem1a:
I think the landing page should feature sign-up over login, reasoning being that the purpose of a landing page is to entice new users to use a service.

Existing users will need to login, but they also will likely never be on the landing page after the first time (like how we almost never go to Github or Discord's landing pages, even though we use them daily). I feel like existing users are more often just navigating directly to the puzzle dashboard or puzzle instance pages and those pages redirect to login when not authenticated, which makes sense for them.

src/components/Global/SignStatusButton/index.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { Button } from '../index';
+
+const SignStatusButton = ({ status }) => {
+  return (
+    status !== 'authenticated' && (
```
Opeyem1a, line 6: This feels so specific for a component, the `'authenticated'` part (the status value in general) depends on logic outside of this file so perhaps we could just conditionally render this button where it's being used?

Opeyem1a: Or maybe just add an extra prop to the `LogStatusButton` component and add this button there?

Opeyem1a (changes requested):

ketphan02:
Draft this until #101 is merged

yuqi88:
changes:
The login button in the header switches between sign-up & log-off.
Redirect the users to the puzzle-map page instead of the home page after logging in.
Added log-off button to Navbar (the header in the puzzle map & puzzle list). Switch between (profile button + img) and (log-off button) depending on if the user is logged in.
improveHeader_map(logged off)
improveHeader_map(logged in)
improveHeader_home(logged off)
improveHeader_home(logged off)

src/components/Global/LogStatusButton/index.tsx
```diff
@@ -1,13 +1,23 @@
...
+import { LogStatusButtonProps } from '../../../types/logStatusButton';
 
-const LogStatusButton = ({ status }) => {
-  return status === 'authenticated' ? (
-    <Button style={'flat'} content={'Logout'} onClick={() => signOut()} />
-  ) : (
-    <Button style={'flat'} content={'Login'} onClick={() => signIn()} />
-  );
+const LogStatusButton = ({ status, usedInPage }: LogStatusButtonProps) => {
+  if (usedInPage === 'home') {
```
Opeyem1a, line 7: The logic should be cleaned up in general, but also why is this a string parameter? Aren't there only 2 options

Logic should be cleaned up so the "logout" button is only written once in the code

Opeyem1a (changes requested):
Almost there, just some minor fixes

src/components/Global/LogStatusButton/index.tsx
ketphan02: I think we can pass a parameter named link and link whatever link we want to and also can be used to check if need signin() on click

Opeyem1a: This feels more complicated, I think let's just stick with the proposal I made to get this merged

src/components/Global/LogStatusButton/index.tsx
```diff
@@ -1,13 +1,22 @@
...
-const LogStatusButton = ({ status }) => {
-  return status === 'authenticated' ? (
-    <Button style={'flat'} content={'Logout'} onClick={() => signOut()} />
-  ) : (
-    <Button style={'flat'} content={'Login'} onClick={() => signIn()} />
-  );
+const LogStatusButton = ({ status, useLogIn }: LogStatusButtonProps) => {
+  if (useLogIn) {
+    return status === 'authenticated' ? (
+      <Button style={'flat'} content={'Logout'} onClick={() => signOut()} />
```
Opeyem1a, line 9: Why duplicate this code?

Only the second button needs to be conditional

Opeyem1a (changes requested):

Opeyem1a (approved):

PR #127, opened 2022-04-05
change hash.js to the argon2 library

yuqi88:
- changed the hash library

The argon2 library was added in PR #101, but the branch is not merged yet.
This needs to be merged so the other tasks related to admin don't produce more issues.

Close: #100


Opeyem1a:
Is this branch still a draft? @yuqi88

yuqi88:
@Opeyem1a How do we delete a PR? We don't need this anymore.

Opeyem1a:
@yuqi88 There should be a close PR button near the bottom?

PR #131, opened 2022-04-12
Dockerization for production

ketphan02:


ketphan02:
The change in `../map/index,.tsx`: the server is down when there are no puzzles.

[DISCUSSION]
Do we need to seed some puzzles for production?

